PR #18, opened 2024-04-18
Monies

cubebotfan:
Rename bountiful files so that the bounty board works again... Hopefully (I didn't test it)


cubebotfan:
Oh nevermind, Cyb already made a better pr for this

PR #44, opened 2024-04-19
Create optional compat

Deepacat:
Removes create processing for the most part, along with certain blocks that could be too strong or overshadow gt machines
Adds optional recipes for create items to gregtech machines
Disables creates worldgen
Hides some categories to clean up EMI


Deepacat:
(Has been testing with and without create, seems to cause no issues without)

Deepacat:
Also this is an initial base to work off, could remove more and potentially gregify the recipes

PR #45, opened 2024-04-19
Update the Genisis chapter

zcarp7220:


config/ironfurnaces-client.toml
```diff
@@ -0,0 +1,277 @@
+
+#Furnace Settings
+[furnaces]
+	# The capacity of the recipe cache, higher values use more memory.
+	# Default: 10
+	#Range: 1 ~ 100
+	recipe_cache = 10
+
+	[furnaces.energy]
```
LetterN, line 9: this looks like serverside configs on client?

zcarp7220: this is unchanged by me

LetterN: oh nvm for some reason ironfurnaces just use server configs

PR #79, opened 2024-04-21
Fix creative chest recipe

JuiceyBeans:
Creative chest recipe now gives 2 chests instead of 1, so that the player can have infinite creative chests once they craft it. Same recipe as in Nomi


LetterN (approved):

ThePansmith:
tbf you can use JEI drag for that but eh

Issue #114, opened 2024-04-25
[Feature]: Add Lootr


JuiceyBeans:
### Suggestion

Would be really nice especially since Nomi (and by extension, Moni) is a pretty popular modpack for servers to run

WithersChat:
I can do that, IDK if Pansmith is okay with it tho.

Issue #102, opened 2024-04-24
[Discussion] Mods to remove/keep


JuiceyBeans:
There are a few mods that could be removed to save on performance in Moni that I found

**Definitely remove:**
- Artan'sProjectCoreMod: Probably a leftover from some removed mod
- Lemon Lib: Used for MiniUtilities, which was removed
- Stack Refill: Functionality already provided by Inventory Tweaks Refoxed
- CIT Resewn: Wait why is this here anyway? Isn't this a Fabric mod?

**Mods that could be removed:**
- AE2 QoL Recipes: All this mod adds is a bunch of recipes, the mod itself is empty otherwise. Could be implemented via KJS instead
- Leaky:  Its unlikely that players will ever build vanilla farms when HNN exists, so its usefulness is incredibly small
- Energy Meter: To be honest, this one can be kept for all that I care, but its a question of if it really adds anything to the pack. Players can add it themselves if they really want to measure their power rates
- 3D Skin Layers: Again, does it really add anything to the pack? Clientside so players who want it can add it themselves
- Showcase Item: Same as above
- Water condenser: Not sure about this one, might be worth looking into if KJS can make a block that outputs water. If so, that would remove the need for this mod entirely
- Ender Crop: With HNN, why is this really requred? Its going to be obsolete so fast and that's assuming the player even obtains it before achieving sim chambers
- Model Gap Fix: Also a client side mod that really adds nothing to the pack. However, I am in favor of keeping this because why is this not in base MC anyway?
- Visual Workbench: Client side, doesn't really add anything
- WaterFrames: Ok yeah, frames are funny for multiplayer. But the mod is like 54MB so is it really worth it? Keep in mind that not many people are going to use it in singleplayer or even in multiplayer for that matter except to goof off
- Wireless Chargers: Not sure how much use this mod will get, Nomi doesn't have any blocks with its functionality either afaik
- Overloaded Armor Bar: What use is this going to have in Moni, a peaceful modpack lol. Plus I don't think there's any actual armors that players will use that needs an overloaded armor bar
- No Recipe Book Reborn: Is the recipe book really that annoying? It only takes up a few pixels on the screen. Plus there will be players (myself included) who use it, even if just to quickcraft from materials available in the inventory
- No Auto Jump: There's going to be SOME people out there who want auto jump. Just ship the pack with it disabled by default in ``options.txt`` instead of adding another mod; this lets the players who want it enable it as well

**Performance mods:**
[...]

PR #115, opened 2024-04-25
Modlist rework (now with zbighuge performance improvements!)

JuiceyBeans:
**Breaking changes:**
- Water condensers will disappear and need to be replaced by water collectors

**Other:**
- Leaky report threshold bumped from 80 to 200 

**Added:**
- Lootr (361276): #114
- Monazite (1015315): #151
- Water Collector (1016115)

**Removed:**
- Artan'sProjectCoreMod (585454): Unused library
- Lemon Lib (471136): Unused library (originally used by MiniUtils)
- Patchouli (306770): Unused library
- TexTrue's Embeddium Options (910506): Included in Embeddium
- Water condenser (663013): Replaced by Water Collector
- Stack Refill (411813): Obsolete, added by Inventory Tweaks Refoxed
- KleeSlabs (241895): Breaks GT hammers
- 3D Skin Layers (521480): Clientside, can be added by players if they really want it; adds nothing to the modpack
- Overloaded Armor Bar (314002): Moni is peaceful and doesn't have many armor sets that go over the vanilla limit
- Saturn (670986) Increases memory usage for some reason; CPU improvements are negligible and the tradeoff is not worth it imo; more in #102
- No Auto Jump (282099): Same as above; removes an accessibility option
- Ender Crop (242269): Obsolete, HNN is faster and usually obtained before an actual ender crop (Technically required for Enderman model)
- AE2 QoL Recipes (907026): All this mod adds is a bunch of recipes, the mod itself is empty otherwise. Implemented via KJS instead
- Energy Meter (532169): Quote from Xefyr0: ``This thing allows for the transfer of up to 1 GFE/t, as opposed to the Draconic Superconductor Conduit's 134MFE/t. If you want conduit progression to have some weight then this is worth removing``

**Feedback required before removing:**
- Showcase Item (627196): Clientside, can be added by players if they really want it; adds nothing to the modpack
- FTB Quests Optimizer: Not required on client, could just pack it with server files instead
- Smooth Boot: Doesn't seem to do much
- Smooth chunk save: Testing required

**Updated:**
image


LetterN:
- keep 551314
- keep 282099 i do not want to reset my `options.txt` every time i git pull/swap branch etc
- keep 242269 early ender pearl same as nomi

JuiceyBeans:
To be noted: Ender pearls can be made by alloy smelting a diamond with pulsating dust (which can be obtained by smelting uraninite dust too) at LV. They also spawn in Lost Cities chests

WithersChat:
> keep 551314 [No Recipe Book Reborn]
Agreed
> keep 282099 [No Auto Jump] i do not want to reset my `options.txt` every time i git pull/swap branch etc
I think we should just get rid of it and let people disable auto-jump. I never feel good about removing an option from Accessibility Settings, y'know?
> keep 242269 [Ender Crop] early ender pearl same as nomi
As Juicy said, 2 other ways to get it. Also Ender Crop literally requires 4 pearls to craft so it doesn't replace anything already done by something else.

Xefyr0:
I would also like the removal of Waystones, or else its replacement with Tempad.

Either way, /home from FTBUtils already exists.

Leclowndu93150:
Keep Forge CIT, i used it to reskin the jetpacks and components because they have NBT

JuiceyBeans:
> I would also like the removal of Waystones, or else its replacement with Tempad.
> 
> Either way, /home from FTBUtils already exists.

Waystones was always meant to be temporary iirc, will talk about Tempad with Pansmith

JuiceyBeans:
### Current state:

Added:
- Lootr

Removed:
- Artan'sProjectCoreMod
- KleeSlabs
- Leaky
- LemonLib
- Overloaded Armor Bar
- Saturn
- Skin Layers 3D
- Stack Refill

JuiceyBeans:
Updated feedback required section, would love some rn

ThePansmith:
Was actually in the middle of typing that lol:

Added:
    Lootr
    Monazite
Keep: 
    WaterFrames (Shitposty images are good for morale :trollface:, could possibly be removed once the Central Monitor gets ported)
    Wireless Chargers (We lack a way to wirelessly charge things)
    WaterCondenser (Sure, the water pump exists, but this is supposed to be a earlygame compact option. I do want to write my own version of this mod where it will autopush water)
Remove:
    Artan'sProjectCoreMod
    KleeSlabs
    Leaky (Could possibly leave the config for this one in, bumped up to like 200 items or something, just to help servers)
    LemonLib
    Overloaded Armor Bar
    Saturn
    Skin Layers 3D
    Stack Refill
    Ender Crop
    No Auto Jump
    AE2 QoL Recipes (Implement as a KJS script before removal)
    Energy Meter

Leclowndu93150:
I can propose to write the Water Collector mod

JuiceyBeans:
Performance seems to have increased not insignificantly after updating mods

SrRapero720:
I reduced the watermedia jar to 26mb (clearly a feature)

PR #119, opened 2024-04-27
tweak create compat

cubebotfan:
Tweaks the create compatability scripts
-Enable more contraption parts like saws and rollers
-Removes the hand crank
-Create recipes use rubber instead of kelp
-Create stone variants now have rock breaking recipes and atomic reconstructor recipes
-Many ore washing and chemical bath recipes can be done with the create fan now
-Added gt recipes for some create items that got missed
-And some other changes that I forgot


LetterN (approved):

cyb0124:
Train track is uncraftable. There is only a sequenced assembly recipe for it that requires a press, which is disabled.

cubebotfan:
> Train track is uncraftable. There is only a sequenced assembly recipe for it that requires a press, which is disabled.

Just made a quick fix by re-enabling the press for the moment until all the create recipes are gregified.

ThePansmith:
See about fixing that then ill merge

PR #120, opened 2024-04-27
generic update

LetterN:
- update buildtools to symlink mod folders now
- update buildtools to allow windows users to easily set curseforge key (`--key=CFKEY`)
- swapped quest links, now circuits/circ assems reference from the progression, required for packswitching shenanigans


kubejs/server_scripts/_packmode.js
```diff
@@ -11,18 +11,34 @@ if (!config || !config.mode) {
...
 global.packmode = packMode;
 global.isNormalMode = packMode == 'normal';
-global.isHardcoreMode = packMode == 'hardcore';
+global.isExpertMode = (packMode == 'expert') || global.isOmegaMode;
 global.isOmegaMode = packMode == 'omega';
 const isNormalMode = packMode == 'normal';
-const isHardcoreMode = packMode == 'hardcore';
-const isOmegaMode = packMode == 'omega';
\ No newline at end of file
+const isExpertMode = (packMode == 'expert') || isOmegaMode;
```
Guvante, line 27: References `isOmegaMode` before defining it, likely you want to swap the lines (maybe reverse all three so it is consistent.

Ditto for `global.isExpertMode` above

LetterN: this is js we can get away with this, although i don't think i cant get away with this on `global.*` since it does some internal conversion

WithersChat: Oh BTW we won't need an "isOmegaMode" thing, because Omega mode will be a different pack in a different repo when the time comes (too many changes).

PR #136, opened 2024-04-30
ExtendedCrafting Monifactory edition groundworks

WithersChat:
Current version of the mod only has the non-auto Epic table and its JEI integration, but having it should allow to define recipes on a 11x11 crafting grid, or requiring a Tier 5 table.


kubejs/client_scripts/JEI.js
```diff
@@ -104,6 +106,10 @@ JEIEvents.hideItems(event => {
 
     //Greg Milk
     event.hide('gtceu:milk')
+    
+    // Posttank stuff
+    event.hide(/gtceu:uxv/)
+    event.hide({/gtceu:opv/)
```
Guvante: There is an unmatched `{` here

WithersChat: Fixed.

PR #138, opened 2024-05-01
Average Bug Fix PR

Blackalf14:
. Change the last craft still using AE2 Certus Quartz to use the gtceu version
. Overwrite the Inscriber recipe for fluix dust which would give the AE2 version (Removing / hide the vanilla recipe failed)
. Hide unused AE2 stuff
. Remove vanilla recipe for Ender Fluid Conduit (not supposed to be here + uncraftable)


Blackalf14:
Will add to my other currently open PR

PR #142, opened 2024-05-01
Soularium alloy smelter recipe

Deepacat:
recipe parity with Nomi Ceu


LetterN (commented):
huh i swore i added this

LetterN (approved):

Issue #154, opened 2024-05-04
[Bug]: Gas Collector doesn't collect Air in the Void


Blackalf14:
### Issue Summary

Gas collector doesn't collect air in the Void while jei says it should

I couldn't find how to add a Gas Collector Recipe with kjs so I let it to you

### Reproduction steps

1. Go in the Void
2. Setup Gas Collector

PR #155, opened 2024-05-04
Numismatic Dynamo quests rewrite

Blackalf14:
Numismatic Dynamo quests rewrite to remove the part about the Lapidary Calibration augment which no longer exists + fix the use of "Hostile Neural Network" in one of the quest for consistency between the two


Blackalf14:
Now fixes #154

PR #167, opened 2024-05-09
Emergency: Fixing Normal Mode (+ Subatomic Digital Assembler)

WithersChat:
- Yeah basically most recipes don't execute in normal mode currently because of a few small mistakes that could be discovered by loading the pack in normal mode once.
- So I fixed them.
- Also contains the Subatomic Digital Assembler. Considering how bad the scripts were broken, I was more focused on making a fix than on cleanly separating my fix from my overall progress.


WithersChat:
Also contains the Subatomic Digital Assembler. Considering how bad the scripts were broken, I was more focused on making a fix than on cleanly separating my fix from my overall progress.

PR #169, opened 2024-05-09
Replace Waystones with Tempad

JuiceyBeans:
Features:
- Tempad uses Moni pennies as fuel. There is no cooldown as such
- Tempads can be found in dungeon chests (including Lost Cities chests) with a 2% spawn chance
- Tempad recipe is currently unchanged

Notes:
- Someone should probably add a quest for Tempads
- Might be worth reducing the number of FTB homes as Tempads allow TPing anywhere endlessly
- Waystones will be kept in the next release as a deprecated item to let players bookmark their existing waystones
- Tempad has built in compat with Waystones so could be useful if we decide to keep Waystones. This will also make migration easier

### DIFF
--- a/config/tempad.jsonc
+++ b/config/tempad.jsonc
@@ -1,0 +1,51 @@
+{
+    /*
+     * The distance from the player that the Tempad is summoned.
+     * Range: 1 - 10
+     */
+    "distanceFromPlayer": 3,
+    /*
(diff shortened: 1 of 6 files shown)


cubebotfan:
Why the hell would we do this?

ThePansmith:
Not really a fan, the extra /homes makes it redundant, and the waystones at the very least allow other players to use it.

PR #220, opened 2024-05-17
fixed nuclearcraft fluid hiding

Deepacat:
Hopefully put all the useful fluids, it should be easy enough to unhide fluids by adding to the list if they're found


Deepacat:
only loads after client reload due to fluid registry happening after client scripts

Deepacat:
wait im trolling

PR #226, opened 2024-05-18
Removing Draconic Evolution

ThePansmith:
- [x] Item Replacement
- [x] Recipe Replacement
- [ ] Crafting Injector Mechanic Replacement


ThePansmith:
Fusion crafter replacement multi will likely require coremodding, so for now the fusion crafter will still be used.

PR #269, opened 2024-05-28
Fixed pack-mode-switcher.sh so that it works

TheRealLynx:
slight changes so that the script actually works


ThePansmith (commented):
Missing option for harder mode

PR #286, opened 2024-06-02
Quest book fix, EMI cleanup, Added RHF and BBC assline recipes with research

Levlion2:
Cleaned up some useless NuclearCraft items from EMI
Cleaned up some useless Enderio fluids from EMI


kubejs/server_scripts/End_Game.js
zcarp7220, line 1: Already done on #274

kdcjxbsdnbgfdg: oh lol. I forgot to push this and in the time it was already done lmfao

PR #299, opened 2024-06-07
Added micro miner mission plus other changes

Levlion2:
Zcarp on a trip so i changed bulk blast chiller and rotary hearth furnace recipes a little for him, he agreed i think
fixed parallel on rock cycle simulator (it was not intended to be left out)
added micro miner mission for a bunch of skibidi ores, nothing too game breaking
added tiered recipes for ae2 components, pattern provider and interface
added recipes: ev 4x, iv 8x, luv 16x for both interface and pattern provider
Feedback would be welcome


ThePansmith:
Most of the changes are fine, but the extra pattern provider recipes seem redundent.

PR #302, opened 2024-06-09
Split Omnium & requisite element quests into its own questbook chapter

Xefyr0:
This pull request only affects the config overrides (The files that are copied into the actual config folder by the pack mode switcher)

It should be noted that the quest chapter files in the config folder don't match those for any mode. (As in, running the pack mode switcher changes the quest files no matter which mode you choose, which is made clearly visible with Git) This is an issue, since most users will likely assume the pack to be in "normal" mode on download, which it isn't right now.


ThePansmith:
The differences between the config folder and normal mode config override folders is simply just me forgetting to paste the changes over. I probably should make a script for that.

As for a possible chapter split, seems a bit redundant.

Xefyr0:
> seems a bit redundant.

Quest count breakdown per chapter (Normal mode):
Genesis: 59
The Beginning: 69
Simulating Resources: 14
Matter Energy: 43
Early game: 77
Into the Microverse: 39
Midgame: 115
Late Game: 94
Fun with fusion: 34
End Game: 183.

Endgame has almost twice the amount of quests as the next highest chapter, and sits at 7 standard deviations above the mean quest count, even when discounting the Simulating Resources chapter as an outlier. It's quite the overweight chapter.

The endgame quests were in the same location as Omnium quests for Omnifactory and Nomifactory since the endgame was an afterthought in those modpacks, but that's not the case for Moni. The revamped post-tank content deserves a dedicated chapter.

PR #308, opened 2024-06-09
fix dilithium_helium recipe

cyb0124:


IamNotBrianZ:
I fixed this few days ago, but pan hasn't approve it.

PR #324, opened 2024-06-13
Cleanroom hatch tweak

WithersChat:
- Now gated behind Nichrome Coils (RTM alloy cables) at EV.
- Should avoid having an obnoxious amount of cleanrooms.
- Won't make cleanrooms obsolete since you still need sterile cleanroom which the hatches can't do.


ThePansmith:
Too early

WithersChat:
That's... the point? Multis in cleanroom has always been bs anyways.

PR #341, opened 2024-06-25
Adding Thermal Spores to Greenhouse & Adding Ender Spores

Midius:
This pull request adds the missing recipes for Thermal Mushroom Spores to the Greenhouse. To maintain game balance, only the boosted versions of the recipes are included.

The removal of the Ender Crop mod has made acquiring Ender Pearls significantly more difficult in HM, which in turn makes producing items like Vibrant Alloy more challenging. By introducing Ender Spores that align with other Thermal spores, this change will restore the balance to be comparable with CEU HM.

.::edit: Also added a fix for Hydrogen Peroxide, didn't want to make a separate PR for that ::.


WithersChat:
All in order aside from the conflicts.

Issue #345, opened 2024-07-02
Posttank cleanup


ThePansmith:
Generally cleaning up post tank progression, now that we have people to test our spitballed numbers lol. These do not have to all be done in the same PR.

- [ ] Fix higher tier circuit recipes to actually be overall cheaper than previous
- [ ] Add a higher tier of recipe for HOTU (maybe monium ingot for a stack or two?)
- [ ] Buff scrapping output chances to be higher (50%?), but make the stages of the scraping itself be like 95% chance of success or something so you still gotta set up infra
- [ ] A better dilithium recipe for posttank
- [ ] Make Creative Energy Come Earlier, allow it to be used in harder.

PR #347, opened 2024-07-03
New dilithium mission

SparkedTheorem:
New dilithium mission per #345.
Also makes Dilithium a real GregTech ore instead of a fake one. Dilithium has doubled yield compared to before, 4 if you oreproc and 2 if you just smelt. The new dilithium mission gives about 2048 crystals per operation, and also gives End Monazite Ore for the endgame as it is used in Dimensional Processing Units.


SparkedTheorem:
temporarily converted to draft as I work out balancing

SparkedTheorem:
ended up deciding the recipe is fine as is

PR #348, opened 2024-07-04
Attempt 2 at making a EIO Clear Glass recipe

Xefyr0:
This time, the recipe is same for all modes (Normal, Hard, Harder)
Both Quicklime and Soda Ash (common glass additives) can be used to provide the player a choice on which recipe they prefer.
1 tiny pile of either dust per Clear Glass because it's a cosmetic block.


WithersChat:
All clear, except that I'd make the recipe shorter ideally but that's a detail.

PR #371, opened 2024-07-14
Remove Abyss Shard, replace with Echo Shard

Xefyr0:
Self-explanatory.
Possible consequences include:
- people needing to make fewer T8MMs before launching a T9MM mission if they explore the Deep Dark.


WithersChat:
Not convinced on this one, having echo shard be a lower tier and abyss shard be higher tier sounds better, kinda like dragon egg/chaos shard in DE

ThePansmith:
agreed with above.

PR #376, opened 2024-07-16
Hadal Core Texture (Based on Sparks' Hadal Shard)

Xefyr0:


ThePansmith:
mfw saw this right after pushing 0.6.1

Xefyr0:
Don't think this matters much until Sparks' shares their work on the shard tbh

PR #377, opened 2024-07-16
Sculk Solar Textures.

Xefyr0:
Bathyal solar uses Cryolobus sides & regular sculk top.
Abyssal solar uses Cryococcus sides & regular sculk top.
Hadal solar uses Cryococcus sides & hadal sculk top (Palette from Sparked's shard spritesheet)


WithersChat:
Just added textures, can be merged as is.

PR #378, opened 2024-07-16
Remove Infused Lava Crystal Textures

Xefyr0:
Missed these when removing ArmorPlus 💀


Xefyr0:
Missed Infused Lava Crystal Block Textures. Will make another PR if someone else doesn't get to it before me.

PR #386, opened 2024-07-18
Adding Ore & Fluid veins to Luna

Midius:
- Removing double entry for void oil
- Adding Deuterium and Helium-3 to Luna
- Porting Luna ore generation from CEU (with some artistic modifications)
- Nerf Thermal Spores, now slightly less OP


Midius:
Because the Ore Gen Offset is set to 0 in the configs it looks really uniform:
grafik

Also adding a Copper vein to the moon, because copper from the end sucks....

PR #398, opened 2024-07-20
Quest Cleanup Attempt

Xefyr0:
Too many changes have been made to Normal Mode but not Hard mode, or to the FTBQuests config folder but not config-overrides.

This is an attempt to propagate those changes to all applicable locations.
With the breadth and complexity of these changes, however, it's **very important** that this PR be reviewed carefully - it's not very clear which differences between NM and HM are intentional or a result of the issue described above.


ThePansmith:
tbh i should add copying the ftbq config folder to nm config overrides, been meaning to but i keep on forgetting lol

PR #406, opened 2024-07-22
KubeJS Item Reordering, Texturing

WithersChat:
- I finished what was started months ago when the Warden data model was introduced.


Xefyr0:
Not a huge fan of removing the Dragon model. It has use before the PR for both the Ultimate Material and Stem Cells.
Also, the Abyssal Energy core is rotated? The center hole is wider than it is tall, the opposite of all GT energy storage items (And the Bathyal Energy Core!)

ThePansmith:
agreed with above regarding dragon model, great work otherwise

PR #428, opened 2024-07-31
Release Workflows

hk21702:
Implements release workflows for GitHub and CurseForge.

Hooks into existing build workflow.
Utilizes automated semantic versioning.
Files are renamed to have the version in them.
Option to autogenerate change logs via pull request names.

Workflow has advanced options to take the artifacts from a previous run instead of making a new build.

**Requires a Curse Forge release token to be set as RELEASE_TOKEN in secrets.**
**Requires CF project id to be set as CF_PROJECT_ID in variables**

Example of a GitHub release (No manual edits or inputs at all):

image

### DIFF
--- a/.github/workflows/build_pr.yml
+++ b/.github/workflows/build_pr.yml
@@ -3,14 +3,24 @@ on:
   workflow_dispatch:
     inputs:
       target:
-        description: 'Build what'
+        description: "Build what"
         required: true
-        default: 'build-client'
+        default: "build-client"
         type: choice
         options:
           - build-client
           - build-server
           - build-all
+  workflow_call:
+    inputs:
+      target:
+        required: true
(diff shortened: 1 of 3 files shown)


hk21702:
Currently work in progress. The GitHub releases are fully functional. Working on CF ones now.

hk21702:
Example of what the upload on CurseForge looks like. I set the change log to just link to the release so there's only one place that needs to be modified if the change log thats auto-generated needs to be changed. 

image

hk21702:
The CF releases uses this action: https://github.com/Xikaro/upload-curseforge-modpack-action
It's not the best action, and I would have preferred to use https://github.com/itsmeow/curseforge-upload but that one doesn't yet support server file uploads.

PR #438, opened 2024-08-05
HM mm recipe fixes & qb updates

Samaritan314:
-Fixed the hardmode microverse mission using wrong ender dragon scales

-Updated the hardmode questbook to keep up with the latest sculk overhaul like the normal mode qb does

-Added a dependency to the t4.5 mm quest to remind players of the need to build a microverse mkII first

-Changed the quest layout a bit to fix some quest overlaps in hm


ThePansmith:
ah fuck i forgot to update the qb, thanks for that!

PR #444, opened 2024-08-06
Better Discharger Multiblock

Xefyr0:


Xefyr0:
Blud changed the multiblock without decreasing the minimum Cryolobus casing requirement 💀
This actually prevents the multiblock from properly forming.

Moreover, why the modifications in the first place?
The more complex design should be too much of an issue given:
- Recipe times that take fractions of a second means most players won't need to make more than one or two
- The modifications only save ~40 casings, a ~20% decrease.
- The Rotary Hearth Furnace has a similar level of complexity

PR #449, opened 2024-08-06
Supercomputer & Superfabricator parallelable

IamNotBrianZ:
I just built these machines, and i think they're too slow 
Buffed them by making them parallelable

To be more specific on how slow it is now.
When overclocked to UHV and running a warden model, it is only 32x speed.
And i already have a HNN array of 100+ simulation chambers back in EV.


Xefyr0:
There are some side effects to this change:
1. The multiblocks have perfect overclock
2. The multiblocks no longer need maintenance

Still a very good idea, though

Xefyr0:
Unless I'm reading the diff incorrectly, it seems these changes also remove the need for maintenance?
This probably makes sense given the fact that HNN multiblocks don't need maintenance either :V

IamNotBrianZ:
Add back maintenance hatch (I'm not sure on this one, personally speaking i don't want this hatch)
Make parallel hatch an optional component
Fix overclock logic (not perfect overclock anymore)

PR #450, opened 2024-08-06
Gregified recipes for Building Gadgets

Xefyr0:
Most gadgets should be accessible in MV, with the exception of the Exchanging Gadget which requires an IV emitter in Hardmode.


ThePansmith:
Seems unneeded, given that EB also exists. 

Though i wouldn't mind exchanging requiring mv in NM.

Xefyr0:
I saw Deepa using BG for easy Assline copy/paste, but I think that outside of that QoL there's no real use for anyone except builders. So yeah, this + Shrink mod turning into optional compats would be a neat idea

PR #451, opened 2024-08-06
Rebalance Sculk Supercon to use Cryococcus instead of Cryolobus.

Xefyr0:
In its current state, there's not much reason to use Cryolobus wires & conduits - It's too easy to directly upgrade them to Sculk Superconductor with just some Nether Stars.
This change gates Sculk Superconductor behind Cryococcus instead, so there's at least a short span of time when Cryolobus is obtained but before a Discharger is built when it makes sense to use regular Cryolobus wires.


WithersChat:
Based idea.

PR #452, opened 2024-08-07
Make Atomic Reconstructor recipe more interesting

Xefyr0:
Adds Circuits & lenses to Atomic Reconstructor recipe


ThePansmith:
Seems unneeded, making it use a lens doesn't make much sense as it doesn't fire a laser ala the AA AR.

PR #453, opened 2024-08-07
Create compat update

Deepacat:
Added back recipes to functional some create machines
Added recipes to some of Creates creative items
Added back recipes to some Create machines that still have disabled function, these are for decor and have a tooltip indicating their lack of use for players
Buffed track recipes
"Added" crafts & additions "support" by removing recipe for the op alternator


ThePansmith:
The rail buff is fine but outside of that i dont really think any of this is needed.

Deepacat:
> The rail buff is fine but outside of that i dont really think any of this is needed.

Having creates creative tank is neat for decor and other stuff, worldshaper is neat too, and the other recipes allow using creates stuff again at a time where they don't overshadow gregs capabilities. I think it's balanced and fair for those that want to use create with the pack

nok-ko:
I would really like this to be merged :(

PR #456, opened 2024-08-08
Add thermal mobdrop recipes for HM uses

Samaritan314:
Added blazerod-like recipes (4 dusts -> 1 rod) for the thermal mobdrops so that the pre-mm nether star recipe is available for HM players.


WithersChat:
How the fuck has this not been merged yet?

PR #459, opened 2024-08-08
Add Mega Alloy Blast Smelter, Universal Circuits and tweak some kjs multiblock machines.

FwFeiVan:
Mega Alloy Blast Smelter is a huge ABF. It can parallel and has a 30% acceleration. It's structure is from GTNH:
18fe792427151223ee0efa11109bab08
0db3b44c3ba61985e398e963e5e34390

And the Universal Circuits can be crafted in packer. These sort of circuits are very useful when you want to change the main circuit you are using.
41232ce45bf4f818ec52d2b343343c42
c55462c1ef88e444228b4f08aa297c77

For the tweaks of kjs machines, I add 1toc for all multiblock machines. 1toc can be open with the code "GTRecipeModifiers.SUBTICK_PARALLEL" and it means, when a recipe is already 1tick, the machine will overclock once to get 2 times of max parallel, it's EU cost and parallel will double.

for example, if I use loot superfabricator to run this recipe, It will become 1t when the tier get to ZPM. 
fe9c407767dcc7104c20dc6d08e923b5

Now I use a 16A UIV energy Hatch to provide energy, and the machine has a OpV max EU/t. the machine will overclock 6 times(from ZPM to OpV) and run 64 parallel. At the same time the EU/t becomes 64 times.
02c9ba1809bd28331a73b03f43e18cb5

I also allow some end game machines to use laser hatch, and change the Naquadah reactor's parallel to (2, 2) overclock. It means the time will be half and EU/t will double in an overclock. It allows you to improve it's efficiency but you also need a super huge energy output hatch to get a high power generation.


WithersChat:
Most of this is very solid, but I'd change a couple things:
- The Mega ABS isn't necessary as a multiblock, could instead add the ABS recipe map to the Rotary Hearth Furnace.
- Remove the universal circuits past UIV. UXV and OPV don't have any circuits, and MAX only has one (which looks much better on the final recipe than a tag indicator)

FwFeiVan:
I reserve the change to Nq reactors because it just don't parallel when I test. And it will be to strong if using 4,4 overclock so I think 2,2 overclock is the most reasonable. I also reserve all machine's 1toc.

Xefyr0:
I, for one, like having more cool multiblocks and would like to have a Mega ABS.
that said, there seems to be four duplicate multiblock registry files - Base, Remote, Local, and Backup?

WithersChat:
> I, for one, like having more cool multiblocks and would like to have a Mega ABS.

The issue is, this design is extremely bland. That's an issue GTNH has with some "mega" versions of multis, that U would like to avoid here.

FwFeiVan:
> > I, for one, like having more cool multiblocks and would like to have a Mega ABS.
> 
> The issue is, this design is extremely bland. That's an issue GTNH has with some "mega" versions of multis, that U would like to avoid here.

The most important part of this are 1toc and universal circuit. Mega ABS is just additional to provide a parallelizable ABS because ABS in GTM is more useful than it in GTNH.

Xefyr0:
Got it.
If it's a matter of uninteresting design that can be resolved pretty easily, though (I like designing multis)

Would this be alright?
mega_abs

ThePansmith:
MABS and UCs seem unneeded, but the kjs multi tweaks are nice

FakeDomi:
> Got it. If it's a matter of uninteresting design that can be resolved pretty easily, though (I like designing multis)
> 
> Would this be alright?

@Xefyr0 Can we maybe turn this into a standalone addition for the pack? Parallel ABS would be very nice to have as the freezer recipe for naquadah based alloys in hot ingot form are really stupid.

PR #460, opened 2024-08-09
Update README.md to include server installation instructions

purebluez:
had to figure this out when making a server so thought it would be nice to make a guide for it 
i dont know if the installation process is the same on windows but id bet it would work


README.md
```diff
@@ -49,6 +49,19 @@ Harder Mode is a modifier for Hard Mode, and takes things a step further by:
 - Enabling Pollution and Hazards
 - Various other nerfs
 
-### Installation:
+## Tutorials
```
JuiceyBeans: This should stay as Installation imo

PR #466, opened 2024-08-09
Removed storage drawers, FT was updated to not require it.

IamNotBrianZ:
Considering storage drawer has been deprecated for a while
And we just entered beta stage, maybe we should remove it


ThePansmith:
Required Dependency for framing templates, an issue has been submitted to the FT repo about it. https://github.com/MaxNeedsSnacks/FramingTemplates/issues/1

PR #469, opened 2024-08-09
Quest unification 2

Xefyr0:
Electric Boogaloo. I've a feeling that routinely performing something like this will be a necessity so long as the quest override system remains as-is 💀


ThePansmith:
If you have a better implementation for the quest override system, do share. The less maintenance we have to do, the better lol.

PR #487, opened 2024-08-10
Enable AE2 channels on Harder

Xefyr0:


Mindstyler:
Wasn't this decided against already a few weeks back? At least I thought so.
Channels don't add any realism and are not even a fun mechanic to have, getting worse the more automation-focused a modpack is, aka this one.
Nomifactory HM doesn't have channels enabled.

How about rather enabling hunger for Harder mode? Now that's something I'd look forward to, so eventually making a Stack-o'-S'more HexecontatessaraS'mingot would actually be something to strive for.

Xefyr0:
This is a reasonable point for Normal Mode and Hard Mode, but EM is about masochism for the sake of "realism" and user enjoyment, plus the fact that AE2 channels are enabled by default.

Mindstyler:
True, that's why I do play EM, but there's absolutely nothing realistic about a computer network with a finite connection limit.

Xefyr0:
There may not be anything realistic about a computer network with a finite connection limit, but there _is_ realism in _finite bandwidth._
I will admit that P2Ps kinda circumvent the realism on that front, though😆
The same config that can disable channels also allows for higher channel limits on cables, up to 4x.
So, normal cables have 32 channels and dense cables have 128 with the above commit.
I think that will encourage some basic cable organization instead of the P2P spam that many people in the Discord dread

ThePansmith:
I may consider having a higher channel limit for em, but default channel config kinda just sucks lol (hell, even gtnh has like hyper cables or whatever that exist for the sole purpose of ignoring channels lol)

PR #500, opened 2024-08-12
[Feature]: Make "Charging Batteries" quest in Genesis chapter accept any LV Battery Buffer

### LINKED ISSUE #499: [Feature]: LV Battery Buffer quest should accept all amperage tiers (similar to LV Energy Converter quest)
### Suggestion

Similar to how the LV Energy Converter quest near the end of the Genesis chapter accepts all amperage tiers I think the pack could be improved by having the adjacent quest for creating a LV Battery Buffer also accept different amperage tiers.

image

The mentioned LV Energy Converter quest which accepts any LV Energy Converter (1A, 4A, 8A, or 16A).

image

The nearby LV Battery Buffer quest which only accepts a 4x LV Battery Buffer.

glektarssza:
Closes #499.

image
image
image
image

### DIFF
--- a/config-overrides/harder/ftbquests/quests/chapters/genesis.snbt
+++ b/config-overrides/harder/ftbquests/quests/chapters/genesis.snbt
@@ -1657,7 +1657,26 @@
 			}]
 			tasks: [{
 				id: "0527876D6287634B"
-				item: "gtceu:lv_battery_buffer_4x"
+				item: {
+					Count: 1b
+					id: "itemfilters:or"
+					tag: {
+						items: [
+							{
+								Count: 1b
+								id: "gtceu:lv_battery_buffer_4x"
+							}
+							{
+								Count: 1b
+								id: "gtceu:lv_battery_buffer_8x"
+							}
+							{
+								Count: 1b
+								id: "gtceu:lv_battery_buffer_16x"
+							}
+						]
+					}
+				}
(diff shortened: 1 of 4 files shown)


Xefyr0:
Fun fact: Your changes are incompatible with how pack modes work.
As a result, using `pack-mode-switcher.bat` to change to _any_ mode (including Normal, which the pack is set to by default) will undo your changes.

To fix this, you have to exactly duplicate the changes you made above to all three different Genesis quest files in the `config-overrides` folder.

Is this implementation of pack modes tedious and error-prone, due to it being vulnerable to PRs from people without this esoteric knowledge?
Yes.
If you have any better ideas _please_ let us know.

glektarssza:
@Xefyr0; apologies. I will fix that this morning.

glektarssza:
I did some thinking about it and there are a few ways I can think of that _could_ resolve the issue you described with the various modes. The issue is none of them are really _better_ than what you're already doing. They're all either far more complex or prone to breaking in some other way that is arguable worse.

Maybe put a `CONTRIBUTING.md` in the repository and mention that mentions some files are duplicated across the various pack modes? That seems the most sensible way to handle this in the short term.

ThePansmith:
Yeah, having a contributing file for things like qb changes, optional compats, etc would be a good idea.

PR #505, opened 2024-08-12
Fix UIV dynamo recipe

IamNotBrianZ:
make it use holmium instead


WithersChat:
We should use holmium instead. It's the UIV-tier wire.

IamNotBrianZ:
Alright, got you

PR #512, opened 2024-08-13
Minor recipe tweaks

Xefyr0:
Remove a bunch of Utilitarian recipes that shortcut over GT recipes
Fix Sand->Dust->Clay block recipes for all modes
New NM recipe to make Rubber Ingots from Sticky Resin without Extractor or Sulfur


ThePansmith:
Overall good, but not a fan of of the rubber ingot recipe, may be better to just move the quest to the lv wiremill quest.

PR #541, opened 2024-08-18
Rework infinite Cobble/Water Sources

Xefyr0:
Reworks Water Collector, Igneous Extruder, Aquaeous Accumulator, GT Infinite Water Cover, and ExAE2 Infinity Water/Cobble Cells.

Water Collector is NM only now, HM has Primitive Water Pump so it's fine

Aquaeous Accumulator, Igneous Extruder recipes are changed to be more like other TE recipes, but have similar gating/costs when compared to the previous recipes

Infinite Water Cover recipe reset to GT default. Now gated to HV for all modes, no longer requires Aquaeous Accumulator + Water Collector nested recipes.

ExAE2 Infinite cell made obtainable once again (or, at the very least, the recipe is now visible in JEI) Recipe was also moved to Assembler and some ingredients switched out for _slightly_ more expensive variants that are still obtainable in the same voltage tier since you only need one of each for each AE2 network.

Because Mana Infused Metal is used in their new recipe, however, the quest link from HV circuits is removed since it's not the last hurdle people will have to overcome to make one.

### DIFF
--- a/config-overrides/hardmode/ftbquests/quests/chapters/the_beginning.snbt
+++ b/config-overrides/hardmode/ftbquests/quests/chapters/the_beginning.snbt
@@ -670,7 +670,7 @@
 			y: -0.5d
 		}
 		{
-			dependencies: ["29A487F0BC43AAF9"]
+			dependencies: ["025543EDB0A4AC22"]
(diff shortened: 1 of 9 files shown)


Xefyr0:
> Over all decent, replace some part (maybe the copper ingot) of TE watergen recipe with eio glass to keep it requiring alloy smelter

Electrical Steel and Resonating Crystal in Redstone Servo: 😶

PR #543, opened 2024-08-19
[Feature] Made MV Energy Converter quest in Early Game questline accept all MV Energy Converters

glektarssza:
This questline file does not appear to be replicated across the different difficulty modes so I don't think this change needs to be applied anywhere else. Please let me know if that's not the case.

See below screenshots for the changes working on my personal server.

image

image

image


glektarssza:
Closed due to resetting my fork. Will open on a separate branch.

PR #564, opened 2024-08-22
Reduced Component Cost of Dimensional Superassembler

SparkedTheorem:
The Controller cost a ridiculous amount of Infinity Ingots, leading to a ~400 HotU cost assuming you had the Furious Catalyst or a whopping 1300 HotUs without it.


WithersChat:
Can you also turn the multiblock from parallel to perfect OC? it's not really worth using otherwise because even the biggest input buses don't have enough room to fit a decent amount of parallel due to assembly line recipes regularly using full stacks in them.

PR #571, opened 2024-08-24
No more Free Cake for Harder Mode

Xefyr0:
- Remove repeatable free Cake quest in EM so players have to craft them
- Remove Cake rewards from Genesis quest in EM so that players start with only a QB


ThePansmith:
The cakes are qol

PR #583, opened 2024-08-25
Parallel Implosion Compressor multiblock. Gated to ZPM

Xefyr0:
Idea:
https://discord.com/channels/914926812948234260/1229854271613436066/1275496334035648564


Xefyr0:
Turns out Nomi CEU had a similar idea and _also_ made a parallel implosion compressor multiblock if that's of interest

PR #604, opened 2024-08-26
Added a singular splash test due to a legendary moment that happened in the discord

WithersChat:
For more context:
https://discord.com/channels/914926812948234260/1229854271613436066/1277471922011246676


ThePansmith:
Quoting on the splash list seems a bit odd, but if we ever implement a GTNH style loading screen with the quotes and whatnot this would be peak

PR #608, opened 2024-08-26
GCyR update + Rebuild modlist

JuiceyBeans:
# WARNING: CURRENTLY UNTESTED!!! DO NOT MERGE!!

Rebuild mod list from scratch to remove all redundant libs + update GCyR yippee yippee yahoo wahoo

PR #612, opened 2024-08-27
[Bug Fix] Remove FTB Quest Optimizer

glektarssza:
It causes quests to take FOREVER to register on servers. Sometimes minutes, even after tweaking the config file as recommended in bigenergy/ftb-quests-optimizer#11 (adjusting delay, disabling entirely, etc). Only removing the mod seems to fix the issue.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -116,11 +116,6 @@
       "required": true
     },
     {
-      "projectID": 912469,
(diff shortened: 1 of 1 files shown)


glektarssza:
Closing in favor of #608.

PR #613, opened 2024-08-27
Post-tank polymer PECA to replace PBI

Xefyr0:
So far, there have been a total of four (4) new materials introduced post-tank. This would be the fifth, and the _only_ one that uses fluids as intermediate products.

This is intentional to both give HM/EM players additional challenge in polymer line to setup, and provide NM players a break from re-doing old polymer lines with the Tank by enabling them to skip a new one entirely. This should improve the perceived value of the tank.
[Inspiration](https://discord.com/channels/914926812948234260/1229929078547550238/1277786496224595988)


Xefyr0:
On second thought - GCYR's Para-Aramid might also have some applications in a similar vein, assuming it isn't also slated for removal :skull:

PR #623, opened 2024-08-29
Misc quests for Genesis + easier GT magnet

JuiceyBeans:
To do (help needed):
- [x] Add quests to overrides for HM / HM^2
- [ ] Fix merge conflicts

What:
- Add GT and Thermal magnet to electromagnet quest
- Add quest for furnace augments, settings copier, and factory augment
- Fix an OR filter name
- Add an easier magnet recipe for normal mode
- Change LV wiremill quest icon from LV motor to LV wiremill
- Replace all occurences of "Bending Machine" with "Bender"


glektarssza:
Maybe we should change the Curios slots a bit too so that instead of a Charm slot and a GTCEu Magnet slot it's a Charm slot and a Magnet slot? Not sure how much effort that is but I think it'd be a lot less confusing.

Current Curious layout:

image

JuiceyBeans:
GTm does this, beyond us I think

glektarssza:
> GTm does this, beyond us I think

Fair point. Disregard, then.

JuiceyBeans:
> I did a quick test and it seems like it is possible to do purely from a data pack via KubeJS.
> 
> image
> 
> image

PR it, would be useful

glektarssza:
> PR it, would be useful

Working on it!

Xefyr0:
@JuiceyBeans you forgot to edit the Genesis quest overrides too 💀

JuiceyBeans:
> @JuiceyBeans you forgot to edit the Genesis quest overrides too 💀

The what nows?

Xefyr0:
The config overrides.
The files that _completely overwrite_ the genesis quest file (and select other chapters) when you switch to any packmode, including Normal (which Moni is in by default)

glektarssza:
> I can't figure out how to add the quests to HM / HM^2 without breaking them, it'd be great if someone else could take a crack at those
> 
> (also can we have HM^2 be the abbreviation for Harder Mode lol)

Breaking them how?

Xefyr0 (changes requested):
1. No need for an Early Game config-override for Normal.
2. The Iron Furnaces and Magnet quests seem to be absent from the Harder Mode config override.
3. See if you can get rid of the NBT/Tag requirement for the magnet quest(s).

JuiceyBeans:
>     3. See if you can get rid of the NBT/Tag requirement for the magnet quest(s).
Already tried it, doesn't work

Xefyr0 (approved):

Xefyr0:
IT IS DONE

JuiceyBeans:
:lets: thanks for resolving conflicts Xefyr, I lost motivation for this PR

PR #625, opened 2024-08-29
[Feature] Support Other Magnet Types in Curios slots

glektarssza:
Still to do:

- [x] Add EnderIO Item Magnets to the supported list
- [x] Fix Curio Slot translation (it's not working yet)


glektarssza:
Translated Curio Slot:

image

Slot accepts Thermal FluxoMagnet now:

image

Slot accepts EnderIO Electromagnet now:

image

PR #631, opened 2024-08-29
Super tank quest typo

Aderixx:
The quest description says the capacity of the tier 1 super tank is 4,000 buckets when it is 8,000


ThePansmith:
Need to make sure questbook related changes are applied to all versions of the qb

Aderixx:
The early game chapter is not in any of the overrides for hard and harder modes

PR #633, opened 2024-08-30
Hide excess ores + GCYR UXV/OpV stuff

### LINKED ISSUE #629: [Feature]: Hide all ore variants excluding stone, netherrack, and endstone from EMI
### Suggestion

Having 19 different ore blocks show up in EMI doesn't just clog it up but also makes it harder to find the base ore variant to buy using Monicoins. It can also make things a bit unintuitive for newcomers to GT to use

I recommend leaving only stone, netherrack and endstone in EMI as netherrack and endstone variants provide 2x output

JuiceyBeans:
Resolves #629


three-fold:
Does this preserve the ores you can obtain from Microminers as well as the ones available for purchase?

PR #655, opened 2024-09-02
Un-screw the config overrides.

Xefyr0:
One of these was my bad, which goes to show how error-prone the whole config override system is :(


Xefyr0:
For the 2nd commit, there's a lot going on but basically FTBQuests automatically removes the typing of number literals (as in, 'Count: 1b' turns into 'Count: 1')  when it automatically saves the quest file(s) upon exiting a world and returning to the title screen.
This would show up as a modification according to git, so the whole commit is just to save that extra 'modification' from appearing due to MC modifying the file.

PR #657, opened 2024-09-02
Pulsating mesh update

Xefyr0:
- Updates Pulsating mesh texture to resemble the Carbon Fiber mesh it's made of
- Moves Carbon Fiber mesh recipe to out of Alloy Smelter and adds small glue requirement [at the behest of PureBluez](https://discord.com/channels/914926812948234260/1229929078547550238/1279267105778307072)


ThePansmith:
Like the texture, not feeling the glue requirement.

Xefyr0:
So be it - I only included in the first place since nearly every recipe in the Chem reactor includes some kind of fluid but there _are_ exceptions (such as the Golden Apple recipe)

PR #660, opened 2024-09-02
Less alloy smelter

Xefyr0:
Move a bunch of miscellaneous recipes out of the Alloy Smelter and into machines that make more sense given the nature of the craft, plus more efficient alternatives for some.

Yet again, this change was suggested by [PureBluez.](https://discord.com/channels/914926812948234260/1229929078547550238/1279267105778307072)


Xefyr0:
One other change this PR does is change the TE flux cell recipe to make more sense, in that the recipe cost now reflects the fact that it's a sidegrade instead of an upgrade.

PR #663, opened 2024-09-03
Alternative recipes + Diamond Lattice rebalancing

Xefyr0:
Adds some alternative recipes for intermediate products that don't have any.

Also includes some re-balancing for Diamond Lattices, making them significantly cheaper with regular crafting - to around 2 T6MM's worth of diamonds per T8MM's worth of Crystal Matrix double plates. There is also a UV Assembler recipe that brings their cost down another 30-ish percent.
This hasn't come into play very much for our testers since NM players get tank shortly after Discharger (And as such they don't need to make very many T10s) but it importantly makes Crystal Matrix ingots cheaper for HM/EM players who won't get tank's creative properties.

### DIFF
--- a/kubejs/server_scripts/End_Game.js
+++ b/kubejs/server_scripts/End_Game.js
@@ -136,6 +136,11 @@ ServerEvents.recipes(event => {
             P: 'gtceu:electrum_flux_plate'
         }
     ).id('redstone_arsenal:materials/flux_plating')
+    event.recipes.gtceu.omnic_forge('kubejs:flux_plating_assembly')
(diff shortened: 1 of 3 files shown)


Xefyr0:
Just want to point out that the diamond lattice recipe change is absolutely something that the players desire:
https://discord.com/channels/914926812948234260/1291570341889445942/1294653625305530438
https://discord.com/channels/914926812948234260/1229854271613436066/1282900345613320242
https://discord.com/channels/914926812948234260/1229854271613436066/1291856962287439914

Xefyr0:
As promised: https://discord.com/channels/914926812948234260/1229854271613436066/1295921583742779495

PR #669, opened 2024-09-05
Redo Atomic Reconstructor v2

Xefyr0:
This time, with field generators instead of lenses!

If there is a reason other than the lenses that this was denied last time, I would like to mention that all other machine recipes (except for this one) have circuits, so please integrate a circuit into the Atomic Reconstructor recipe somehow.


ThePansmith:
Overall good, but issue in it needing a field gen. Mainly just cause that LV version exists as qol for builders (create compat) and such (rotten flesh, etc), and having it require MnP feels a bit off.

PR #675, opened 2024-09-07
Add the ReHooked mod

OE100:
The ReHooked mod is my attempt to bring the Hooked mod (that was on nomi ceu) to 1.20.1, I'm still actively working on it but it's at a stable version and all the features of the Hooked mod are already in.

If you're interested in considering it, you can check it out here:
https://www.curseforge.com/minecraft/mc-mods/rehooked

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -971,7 +971,7 @@
       "required": true
     },
     {
-      "fileID": 5701875,
+      "fileID": 5731647,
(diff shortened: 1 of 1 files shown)


ThePansmith:
Didn't feeling like screwing up the conflict resolution given the change in modlist since this PR, so I just did it myself lol. Thanks for the mod!

PR #683, opened 2024-09-10
Buff GT boilers to be almost reasonable compared to the Systeams stirling boiler

Xefyr0:
Fun fact: GT boilers burn fuel at a rate that's 12 (HP) or 24 (LP) times slower than TE Stirling boilers.
However, they also produce steam 10 (HP) or 26 (LP) times slower.
Tripling their steam output will result in GT boilers that produce steam 3 (HP) or 8 (LP) times slower, but are about 3x as efficient on fuel and 16x more efficient on water. Neat.

This PR also has some config updates for Thermal-related stuff lumped in.

### DIFF
--- a/config/gtceu.yaml
+++ b/config/gtceu.yaml
@@ -280,7 +280,7 @@ machines:
 
     # The amount of steam a High Pressure Steam Solid Boiler produces per second at max temperature.
     # Default: 300
-    hpSolidBoilerBaseOutput: 1800
+    hpSolidBoilerBaseOutput: 900
 
     # The amount of steam a Steam Liquid Boiler produces per second at max temperature.
     # Default: 240
--- a/config/gtceu.yaml
+++ b/config/gtceu.yaml
@@ -276,27 +276,27 @@ machines:
   smallBoilers:
     # The amount of steam a Steam Solid Boiler produces per second at max temperature.
     # Default: 120
-    solidBoilerBaseOutput: 120
+    solidBoilerBaseOutput: 240
 
     # The amount of steam a High Pressure Steam Solid Boiler produces per second at max temperature.
     # Default: 300
-    hpSolidBoilerBaseOutput: 300
+    hpSolidBoilerBaseOutput: 600
 
     # The amount of steam a Steam Liquid Boiler produces per second at max temperature.
     # Default: 240
-    liquidBoilerBaseOutput: 240
+    liquidBoilerBaseOutput: 480
 
     # The amount of steam a High Pressure Steam Liquid Boiler produces per second at max temperature.
     # Default: 600
-    hpLiquidBoilerBaseOutput: 600
+    hpLiquidBoilerBaseOutput: 1200
 
     # The amount of steam a Steam Solar Boiler produces per second at max temperature.
     # Default: 120
-    solarBoilerBaseOutput: 120
+    solarBoilerBaseOutput: 160
 
     # The amount of steam a High Pressure Steam Solar Boiler produces per second at max temperature.
     # Default: 360
-    hpSolarBoilerBaseOutput: 360
+    hpSolarBoilerBaseOutput: 480
 
   # Large Steam Boiler Options
   largeBoilers:
--- a/config/gtceu.yaml
+++ b/config/gtceu.yaml
@@ -276,27 +276,27 @@ machines:
   smallBoilers:
     # The amount of steam a Steam Solid Boiler produces per second at max temperature.
     # Default: 120
-    solidBoilerBaseOutput: 240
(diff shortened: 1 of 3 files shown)


Xefyr0:
One random thing of note: the Thermal Expansion Stirling Boiler is the only TE boiler available in HM/EM, in case this was not intentional.

Ghezra:
Would you also consider buffs/changes to the Large boilers as well?

On hard mode a single coal:
```
Boiler:  total mb:   burn ticks:     mb/t:
LP          115200         19200         6
HP          288000         19200        15
LarBronz     16000            20       800
LarSteel     16000            20      1800
LarTitan     36000            20      3200
LarTung      64000            20      6400
Stirling     24000          1200        20
``` 

Its even more egregious when talking about creosote: (per 1000mb efficiency wise)
```
Boiler:  total mb:   burn ticks:     mb/t:
LP          345600         28800        12
HP          864000         28800        30
LarBronz     50000            20       800
LarSteel    112500            20      1800
LarTitan    200000            20      3200
LarTung     400000            20      6400
20x LP      345600          1440       240
20x HP      864000          1440       600
``` 
It takes 240 steel to make 20 HP steam liquid boilers compared to the 138 steel needed for the large steel boiler (which is HV tier), but you'll make ~8x's the steam (at 1/3rd the mb/t) with only being in the Bronze/Steam age versus *HV*.

Not to mention the complete uselessness of the Large Bronze Boiler which makes 1/20th the total steam at only 30% more output.

If you double the mb/t of the small boilers, you make it to where 20 HP boilers now produce 16x's the total steam as a steel boiler, while only being 2/3rd the output rate. And the Large Bronze boiler should just be removed from the game entirely,
Making 1/40th the total steam at 2/3rd the output rate.

I'd like there to be a reason to move to the Large Bronze Boiler and future boilers versus being significantly more efficient to just keep stacking HP liquid boilers being fed creosote.

Xefyr0:
> Boiler:  total mb:   burn ticks:     mb/t:
> Stirling     24000          1200        20

Where did you get these numbers?
A Stirling boiler outputs 192000 mB with a single coal before this commit, not 24k
although I suppose that that makes this PR all the more important :trollface:

Xefyr0:
The real solution isn't to buff GT Boilers, but to nerf TE boilers actually I think
But if we do that then the holy ratio of 1 boiler to 2 turbines would be ruined 😭
IDK I guess I have to consult pan on this? Maybe make TE dynamos produce steam at the same rate but burn through coal much faster

Xefyr0:
yeah no tripling the steam output of the large boilers to match the smaller boilers ups their max temperatures, meaning that the buffed Tungstensteel boiler at full tilt beats out the end-game coil by 4474 Kelvin :trollface:

Xefyr0:
Yea this leaves us with a few options:
- Buff GT boilers to match TE boilers (Insane behaviour, see Large Tungstensteel Boiler having the power of the sun above)
- Nerf TE boilers to match GT boilers (And run the risk of completely screwing up all earlygame power balance)
- Make it so that NM cannot use GT boilers (Limited to TE only) and HM/EM cannot use TE boilers (Limited to GT only)

Closing the PR for now until a decision is made

Ghezra:
>Where did you get these numbers?
A Stirling boiler outputs 192000 mB with a single coal before this commit, not 24k
Pretty sure it's because TE Stirling Boiler produces 160 mB/t instead of 20
although I suppose that that makes this PR all the more important :trollface:

@Xefyr0 
I tested these numbers with a stirling boiler and an empty tank up against its output in creative mode.
An unmodified/upgraded stirling boiler only does 20 mb/t. You have to use augments and upgrades to increase its output from that. (But I'm setup for/am playing in Hardmode, not Normal, not sure if these get affected normal->hard.)

Doesn't change the fact that the Large Steam Boilers pre-titanium are basically worthless compared to HP single blocks.

Xefyr0:
@Ghezra 
Hm. I tested my 192k steam-per-coal number on both Normal and Hard mode and it got the same value both times - are you sure you provided the Stirling/Steam Boiler enough water?

as for the large boilers being "basically useless," you can see in your own charts that they trade efficiency for throughput.
If you want to use the singleblock boilers instead of equivalent large variants, then you'll have to build 60 (liquid fuel) or 120 (solid fuel) of them to match the steam production speed.

With that in mind if you want to continue the conversation, please do so in #735

Xefyr0:
Actually, looks like it [may be a GTM thing too](https://discord.com/channels/701354865217110096/1086528125711700028/1287305857117458502)

So it will probably be fixed eventually.
(On that note, I probably should reduce the amount of nerfing #735 does)

PR #704, opened 2024-09-16
Enable Creative Storage Data to provide Quantum Chest/Tank X in HM/EM too

Xefyr0:


ThePansmith:
Can be shredded for free resources in HM+

Xefyr0:
Huh?
Not on my instance.
The Quantum Chest X is UEV and thus has no recycling recipe, unlike the Quantum Chest IX which is UHV and thus _can_ be recycled for Neutronium.
The Data was changed to produce QC X's a while back instead of QC IX's in #430

PR #710, opened 2024-09-17
Added Quark mod

Deepacat:
Added Quark mod
    - Disabled some useless features such as mobs, loot table stuff, and newly generated blocks.
    - Kept the Shiba though, it's cute
    
Added Zeta mod, a Quark dependency

Removed "Item Showcase" mod, Quark replaces functionality.

commit 2:
Re-enabled variants
disabled normal animal variants because people seem to not like those for some reason
made all bees gay.
kept shinies enabled

### DIFF
--- a/config/quark-common.toml
+++ b/config/quark-common.toml
@@ -1,0 +1,1950 @@
+
+[general]
+	"Enable 'q' Button" = false
+	"'q' Button on the Right" = false
+	"Disable Q Menu Effects" = false
+	#How many advancements deep you can see in the advancement screen. Vanilla is 2.
+	#Allowed values: (0,)
+	"Advancement Visibility Depth" = 2
+	#Blocks that Quark should treat as Shulker Boxes.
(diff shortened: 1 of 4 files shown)


Deepacat:
Yes I ensured it works on previously generated worlds

config/quark-common.toml
```diff
@@ -0,0 +1,1950 @@
...
+	[client.variant_animal_textures]
+		"Enable Cow" = false
+		"Enable Pig" = false
+		"Enable Chicken" = false
+		"Enable Shiny Rabbit" = true
+		"Enable Shiny Llama" = true
+		"Enable Shiny Dolphin" = true
+		"Enable Shiny Slime" = true
+		"Enable L G B T Bees" = true
+		"Every Bee Is L G B T" = true
```
juh9870: Maybe keep this off? Can bump shiny chance way up (it controls those, right? I may be wrong) if you want to see those bees often, but making all of them this way is way too jarring and would make getting regular looking bees impossible for those who want them normal

Deepacat: Shiny chance is seperate but can be increased, I think the bees are fine

Issue #534, opened 2024-08-16
[Bug]: Spelling error on Recipe Trees and Crafting Trees quest


hamburger5003:
### Issue Summary

For the quest 'Recipe Trees and Crafting Trees' in the Genesis tab, the highlighted text says 'Synethic Favorites' and it should say 'Synthetic Favorites'

![2024-08-16_09 06 39](https://github.com/user-attachments/assets/6142ebf3-eaa7-4af2-88c6-730a5f87278c)
![2024-08-16_09 06 39-copy](https://github.com/user-attachments/assets/d7c8394c-2dfa-4747-869d-d31bc6327a05)

### Reproduction steps

N/A

PR #713, opened 2024-09-18
fix typo in recipe trees and crafting trees quest

TNMYP:
#534


ThePansmith:
Don't forget to also commit these changes to the config-overrides versions as well.

PR #717, opened 2024-09-19
Parallel fusion & multiblock improvements

Xefyr0:
I'm just lumping #716 into this one because Git decided to screw me over @ 3 am. Not dealing with it now.
If you for some reason don't like one or the other, the commits are pretty separate.


ThePansmith:
i mean at this point we gotta do them all :trollface:

PR #724, opened 2024-09-20
Nerf Neutronium Solars

Xefyr0:
Neutronium solars are such a large power spike that Infinity solars hit some sort of internal cap Solar Flux Reborn has on how much power they can generate and therefore feel weak in comparison. So, Neutronium solars now only generate 1/4 the power amount they used to.

[This has also been](https://discord.com/channels/914926812948234260/1229929078547550238/1240078546995777619) [a long time coming](https://discord.com/channels/914926812948234260/1229929078547550238/1258944442942095390)


richie3366:
Don't forget to also update the crafting recipes that allows to convert them into GT solar covers

Xefyr0:
Done 👍

ThePansmith:
Fine with me, also note that solars can be backlisted from working in GT tick accelerators

PR #735, opened 2024-09-21
Nerf the fuel efficiency of Thermal's Steam Boiler

Xefyr0:
After a long night of work, this PR would make Steam Boilers 4x as fuel-hungry without changing their steam output.
It also buffs the speed of other Systeams boiler types, since they were pretty slow & underused anyways.

Why do this? See:
https://discord.com/channels/914926812948234260/1229929078547550238/1286931800207261749
https://discord.com/channels/914926812948234260/1229929078547550238/1286941927895339039

(If you don't feel like looking, the tl;dr is that TE Steam Boilers completely outclass any GT alternative and tended to make earlygame power too easy. Also buffing GT steam boilers had some weird consequences)

### DIFF
--- a/defaultconfigs/systeams-server.toml
+++ b/defaultconfigs/systeams-server.toml
@@ -12,14 +12,14 @@ replace_dynamo_augment_tooltips = true
 
 #For future reference: Boilers are 16x as efficient as Dynamos by default.
 #This means that Numismatic Boilers are 16 * 0.5 = 8x as efficient as Numismatic Dynamos,
-#While Stirling Boilers are 16 * 0.125 = 2x as efficient as Stirling Dynamos.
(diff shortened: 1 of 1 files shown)


Xefyr0:
Bringing over some data from #683, I'd like to repeat:
LP Steam Boilers have an output of 6 mB/t @ 115.2 Steam Buckets per coal
HP Steam Boilers have an output of 15 mB/t @ 288 Steam Buckets per coal
TE Steam Boilers have an output of 160 mB/t @ 192 Steam Buckets per coal
the GT LBB has an output of 800 mB/t @ 50 Steam Buckets per coal

This doesn't even account for augments or Integral Components. TE Steam Boilers need a nerf.

ThePansmith:
Meh, wouldn't really do much other than slow down early game. Do keep the buffs for the other boilers, though.

PR #741, opened 2024-09-22
Remove the Pulsating Crystal cost for Snad on NM

Xefyr0:
In response to [backlash on making the Snad recipe harder than seen in Threefold's video](https://discord.com/channels/914926812948234260/1229854271613436066/1287268528063254640)

### DIFF
--- a/kubejs/server_scripts/random_recipes.js
+++ b/kubejs/server_scripts/random_recipes.js
@@ -9,7 +9,7 @@ ServerEvents.recipes(event => {
 
         //If Snad is obtainable pre-autoclave, so must be the Vacuum Chest.
         //Otherwise, people will face laggy items everywhere on the ground.
-        event.replaceInput({ id: 'enderio:vacuum_chest'}, 'enderio:pulsating_crystal', '#forge:gems/ender_pearl')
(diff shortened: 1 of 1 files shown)


ThePansmith:
If you do that, you also need make the item collection options cheaper. Half the reason nomi (and thus moni) have this nerf in place is so that players can craft a item collector before snad, thus ensuring that they cant accidently lag out the game with hundreds of dropped items.

klaborda:
Just curious (mostly since I just started my playthrough in NM and I want this behavior), are you thinking of making the collector upgrade not require the vacuum chest then? I'm at the point in my playthrough where this upgrade would definitely help me out if it didn't require the pulsating crystal

https://github.com/ThePansmith/Monifactory/blob/8499b0c21fe40d857550c797d7e62869f03c14c9/kubejs/server_scripts/mods/functionalstorage.js#L109

Xefyr0:
> Half the reason nomi (and thus moni) have this nerf in place is so that players can craft a item collector before snad, thus ensuring that they cant accidently lag out the game with hundreds of dropped items.

Great! I knew that there was a good reason things were the way they were.

> If you do that, you also need make the item collection options cheaper.

Ok, done. Now the most expensive thing in a Vacuum Chest is Tin - Ender Pearls are pretty pricey in Peaceful (They require alloy smelting) but Pulsating Dust could be another alternative to a Rotor if that change sounds better than my idea

PR #755, opened 2024-09-23
Move some Genesis quests to The Beginning to be compatible with Peaceful playthroughs

inigmatus:
Peaceful players can't finish all Genesis quests.

- Lost Cities and Registering Homes can't be completed until mob learning.
- Electromagnet can't be completed until autoclave.
- Inspector Gadget can't be completed until alloy smelter.

 Submitted Changes:
- Removed Lost Cities entirely and just keep Registering Homes since cake method of overworld transport is ubiquitous from the first Genesis quest.
- Moved Electromagnet to The Beginning and require LV Autoclave quest.
- Moved Inspector Gadget to The Beginning and require Pulsating Iron and Ender Pearls quest.

Note:
This change will require updating the config override files accordingly... but im not sure how to create those on my end. I merely just used the editor in my Peaceful server and saved and downloaded the changes.

### DIFF
--- a/config/ftbquests/quests/chapters/genesis.snbt
+++ b/config/ftbquests/quests/chapters/genesis.snbt
@@ -1610,63 +1610,6 @@
 			y: 15.5d
 		}
 		{
-			dependencies: ["06E9B9DDDB245F38"]
-			description: [
-				"These devices can help you build, swap blocks, destroy massive areas, or even copy-paste regions. You need the blocks you plan to paste in your inventory though; it doesn't make them from nothing."
-				""
-				"Also important to note that it won't work with TileEntities, so it pretty much only works with simple blocks (not stuff like machines)."
-				""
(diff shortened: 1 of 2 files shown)


inigmatus:
Also this change will require updating the config overrides to match... but im not sure how to do those.

Finally, I do suggest renaming Genesis to The Basics simply because "Genesis" is a word synonymous with "The Beginning". and if you are into renaming, I'd suggest changing "The Beginning" to "Factory Machines" so it looks like this: https://i.imgur.com/vrip5pB.png

inigmatus:
Also might be related. But the Lost Cities Cake appears to not be a craftable item. Is this by design? https://i.imgur.com/RagnSjp.png

inigmatus:
Xefyr said he supports adding a recipe for the Lost Cities Cake and making the original Lost Cities method with the bed and skulls a secret route (as in, still enabled but undocumented). If so, then the cake could replace the bed in the Lost Cities quest and still have that quest retained.

inigmatus:
Closing for now. I'm redeveloping this.

PR #770, opened 2024-09-25
Make GT Credits burnable in Numis to match Thermal Expansion coins

Xefyr0:
Also replaces Slime Models' Platinum source from coins to nuggets - this makes the use [more clear](https://discord.com/channels/914926812948234260/1229854271613436066/1288204595419676705) and more [consistent with Thermal's current recipes](https://discord.com/channels/914926812948234260/1229854271613436066/1286837926478741564). The amount of Platinum per matter is not affected - 48mB both before and after.

### DIFF
--- a/kubejs/data/hostilenetworks/data_models/slime.json
+++ b/kubejs/data/hostilenetworks/data_models/slime.json
@@ -32,8 +32,8 @@
             "count": 4
         },
         {
-            "item": "gtceu:platinum_credit",
-            "count": 1
+            "item": "gtceu:platinum_nugget",
+            "count": 3
         }
     ]
 }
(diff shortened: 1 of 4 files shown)


OE100:
should probably leave the recipe for platinum coin -> liquid platinum so that everyone who currently has plat coins stocked up wouldn't be hurt by that.
either that or add a recipe for plat coin -> 3 plat nuggets.

Xefyr0:
Not a good idea - Coin denomination conversion recipes would enable people to transmute _any coin material_ to Platinum.

ThePansmith:
Not really a fan of the whole GT credits bit, but im fine with changing the slime stuff from coins to nuggets.

Issue #787, opened 2024-09-26
[Feature]: Look at reverting gregification of building gadgets for NM


klaborda:
### Suggestion

Explore reverting Xefry0's gregification of building gadgets from [this PR](https://github.com/ThePansmith/Monifactory/pull/699).  Ultimately, this request is driven by the conversation from [this PR](https://github.com/ThePansmith/Monifactory/pull/775#issuecomment-2377632269).

PR #775, opened 2024-09-25
Genesis Peaceful Playthrough Compatibility

inigmatus:
Currently it is not possible for peaceful players to complete Genesis quests in Genesis alone. To fix this entirely:

Universal for all modes: Normal, Hard, and Harder:
- Added craftable Lost City Cake recipe for transport to the Lost City. Refill requires diamond.
- Updated Emergency Items, and rechargeables to include Lost City Cake for Dimensional Transport.
- Updated quest descriptions for Genesis, and Cake Based Dimensional Transit accordingly.
- Renamed Lost Cities and Registering Homes to just Registering Homes and removed bed requirement and reward, as Lost City access is now  accessible from the start via Lost City Cake. 
- Moved Electromagnet to The Beginning with Autoclave requirement since current recipe demands it.
- Moved Inspector Gadget to Early Game with LV Emitter and HV Power requirements since current recipes demand it.

Tested all the above without issues in standard, normal, hard, and harder.

PR #788, opened 2024-09-27
chore: revert gregification of most building gadgets

klaborda:
https://github.com/ThePansmith/Monifactory/issues/787

Reverting gregification of the building gadgets as a follow up to [this conversation](https://github.com/ThePansmith/Monifactory/pull/775#issuecomment-2377632269).  Since EB is back in Moni, there is less of a need to gregificate the building gadget recipes.  With that said, I am leaving the HM recipe modification untouched.

### DIFF
--- a/kubejs/server_scripts/mods/buildinggadgets2.js
+++ b/kubejs/server_scripts/mods/buildinggadgets2.js
@@ -1,104 +1,19 @@
-ServerEvents.recipes( event => {
-    event.remove({id: /buildinggadgets2/})
-
-    //Template Manager
-    event.shaped(
-        'buildinggadgets2:template_manager', [
-            'PMP',
-            'CGC',
-            'PLP'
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/mods/buildinggadgets2.js
```diff
@@ -1,104 +1,19 @@
-ServerEvents.recipes( event => {
-    event.remove({id: /buildinggadgets2/})
-
-    //Template Manager
-    event.shaped(
```
klaborda, line 5: >  I'm fine with it being reverted to default recipes (except for the ones touched by HM recipe changes)

Trying to build off of @ThePansmith 's comment, this is reverting everything except the specific recipe change for HM.  With that said, does this sound right to y'all? Or is this too easy for HM?

Issue #519, opened 2024-08-13
[Feature]: Add a recipe for Calcite in rock breakers


srdr2k3:
### Suggestion

It's a nice looking block and a part of our block palette. It would be cool to see a way to make it renewable.

OE100:
I can't find a reason to not have an hv rock breaker recipe for calcite, I'll try and get it pr'd

PR #790, opened 2024-09-27
added calcite rock breaker recipe

OE100:
exactly what it says on the tin.

the recipe has equivalent requirements and costs as the granite recipe


OE100:
#519

PR #791, opened 2024-09-27
added recipe for sophisticated storage controller

OE100:
only a bit more expensive than the functional storage controller since it's a bit more useful in the early game


Spicierspace153:
honestly, I would say add in a emitter I know its  a bit more expensive but it makes sense for what you use it for if you know what i mean. but thats just my opinion

OE100:
it doesn't really make sense making it *much* more expensive when we have the storage controller from functional storage being so relatively cheap, I say either make both expensive or none.
that being said since this storage controller is more useful for it's stage of the game replacing iron plates with steel (generic slightly more expensive material) makes sense.

we could add an emitter to the remote connector thingy recipe that let's you remotely connect storage blocks to the controller though, because they do add a functionality that could significantly reduce the amount of me storage busses needed early game.

PR #792, opened 2024-09-27
removes Machine press mode from Machine press. replaces with forming

Spicierspace153:
ngl having one mode for a machine is stupid. sorry @WithersChat  but this makes way more sense from a game design for a machine that already had 4 modes.


Spicierspace153:
i was at home eating cool ranch dorito
message ping on discord

"material press mode is kil" 
"yes"

PR #793, opened 2024-09-27
fix circuit on dark steel machine hull

Spicierspace153:
gtm changed the circuit on lv casings causing problems with the maintanance hatch changes circuit to 6 because 6 is what casings are on


Deepacat:
👏

PR #795, opened 2024-09-27
Texture update

WithersChat:
- Updating textures to GTm
- Whoever redid the Neutron Star did an amazing job. I don't even need to retouch it.
- Whoever did Infinity+Monium plate, however, didn't.


WithersChat:
Enjoy :3

PR #796, opened 2024-09-27
Revert "chore: revert gregification of most building gadgets"

inigmatus:
Reverts ThePansmith/Monifactory#788

The initial revert actually reverted the balance changes introduced by @Xefyr0 and incorporated into the quest updates in #775. This pull needs merged before #775 can be merged. @klaborda concurs that the revert should be reverted now to make the intent of #775 possible.


klaborda:
Thank you for submitting this!

PR #798, opened 2024-09-27
Added recipes for building materials

OE100:
Added missing recipes for stone types from quark for the rock breaker and rock cycle simulator.

Replaced the shapeless recipe for tuff with recipes for rock breaker and rock cycle simulator.

Added missing recipes for the polished variants of quark stones.


Xefyr0 (changes requested):
KJS recipes for quark bricks obviated by #823, remove them pls

OE100:
does that resolve all the shaped recipes or just the tuff bricks?

Xefyr0:
No, the previous commit fixed recipes for _all_ quark bricks, not just Tuff.
It's been handled, though.

PR #802, opened 2024-09-27
Added the movin' around quest equivalent from nomi ceu

OE100:
image
image


ThePansmith:
Good, but need to resolve the merge conflicts.

Also, move the applicable quest to the side, looks a bit better.
image

OE100:
done

PR #811, opened 2024-09-29
Fix reference to Snad for RecipeRemoval

mathisto:
I am not certain of the intent, but it looks like this might just be a small mistake. I believe this was meant to remove the easy version of the Snad recipe.


mathisto:
Already fixed in latest update.

PR #814, opened 2024-09-29
Add ModernFix config

Radk6:
Enables `mixin.perf.dynamic_resources` and `mixin.perf.faster_item_rendering` which should decrease RAM usage and loading times as well as increase fps in some scenarios.


Xefyr0:
@Radk6 when you launch the game, a bunch of comments are added to the file to match the mod's default config requirements. can you add the modified file so Git doesn't think I'm trying to make changes when I launch the game?

Radk6:
Should be good now

PR #845, opened 2024-10-04
Quest Chapters

inigmatus:
Based on group consensus, this is this most popular rename since Genesis and The Beginning are synonymous and often confusing. So renaming The Beginning to The Factory makes the most sense. File name changed accordingly as well.

Quest Chapters created and quests reordered in logical progression. Additional lines are moved to Guides & Help. Tested in normal, hard, and harder modes. 
w9gsqJr 1

Genesis is displayed in the Quest Book as:
Normal: Genesis
Hard Mode: Genesis - Hard
Harder Mode: Genesis - Expert

### DIFF
--- a/config-overrides/harder/ftbquests/quests/chapter_groups.snbt
+++ b/config-overrides/harder/ftbquests/quests/chapter_groups.snbt
@@ -1,0 +1,6 @@
+{
+	chapter_groups: [
+		{ id: "5CF31559080744DB", title: "&9Main Progression&9" }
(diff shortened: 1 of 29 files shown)


inigmatus:
I am closing this PR since in examining the whole pack and quest lines, you have actually done a fine job keeping everything mostly where it should be.  Everything but Processing Lines and the Progression quests are Main Progression. 

I will instead close this PR and resubmit a new one with this in mind. Closing this PR.

PR #873, opened 2024-10-07
Updated old Quest descriptions

KePiadina:
Small PR to fix quests using oreDict in the Filters quest despite it being deprecated in favor of Tags, and removing the Nomi GTCEu subtitle as Moni doesn't have Trinium coils as its icon


Possseidon:
While you're at it; the quest states, that the tag filter makes use of regex, but it's actually just some very basic wildcarding with some boolean logic sprinkled on top. Maybe that can be rephrased a bit. The whole syntax is also documented on the filter GUI itself; although funnily enough that documentation is also a bit misleading, since it also uses some weird uppercase examples that don't make sense for tags which are always `snake_case`.

More importantly, does e.g. `forge:hot_ingots` in the filter actually work for you? I couldn't get it to work with the `forge:` namespace. For me it only works if I just do `hot_ingots`. This then basically implicitly wildcards the namespace and would also match items with e.g. `minecraft:hot_ingots` or `foo:hot_ingots` (which is not ideal, but and issue with gregtech rather than monifactory).

KePiadina:
I have tested with ExtendedAE Tag Export Buses, GT Filters, and GT Covers + Filter, and they all work differently.

- Tag Export Bus ALWAYS requires the `forge` tag or wildcard
- GT Filter as Cover works ONLY without the tag
- GT Cover + Filter works with and without the tag

So, for simplicity's sake, I kept it with the tag so as not to confuse players trying to apply filter mechanics to AE2, since they both use Tags in their name.
I will however fix the regex in the descriptions and add a simpler example since tag namespace is not relevant

PR #875, opened 2024-10-07
Removed Quests Steam Grinder, Foundry, & Steam Dynamo from The Factor…

StoneLegion:
…y Harder Mode

Steam Foundry and Grinder are duplicates from Genesis on Harder mode and the Steam Dynamo quests are not craftable in harder mode. All 3 quests are not relevant.

### DIFF
--- a/config-overrides/harder/ftbquests/quests/chapters/the_factory.snbt
+++ b/config-overrides/harder/ftbquests/quests/chapters/the_factory.snbt
@@ -709,20 +709,6 @@
 			y: -0.5d
 		}
 		{
-			dependencies: ["29A487F0BC43AAF9"]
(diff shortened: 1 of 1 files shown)


config-overrides/harder/ftbquests/quests/chapters/the_factory.snbt
```diff
@@ -708,20 +708,6 @@
...
 		}
-		{
-			dependencies: ["29A487F0BC43AAF9"]
-			description: ["Need &9RF&r to charge your items? Thermal dynamos are disabled in harder mode, but you can plug GTEU straight into a &6Tinker's Workbench&r to power it. Such convinence right?"]
-			icon: "systeams:steam_dynamo"
-			id: "545CCC24D9401794"
-			tasks: [{
-				id: "114F684B1AB140C8"
-				title: "Yay RF!"
-				type: "checkmark"
```
TheDogOfChaos, line 719: it's a checkbox task

and it also directly states that 'Thermal dynamos are disabled in harder mode'

PR #879, opened 2024-10-08
AE2 Recipe Tweaks

Xefyr0:
There is only one significant nerf here, and that is to make Crafting CPU housings require 2 Sky Stone Dust each [as an attempt to discourage overuse of autocrafting](https://discord.com/channels/914926812948234260/1229929078547550238/1292876994891219099)

Everything else corrects an inconsistency: 
1. Megacells' storage components didn't follow the circuit tiering pattern set by the AE2 storage components (Increasing circuit tier every two storage component tiers as opposed to AE2's 1 per tier) 
2. When disassembling MAE2's multi co-processors by holding them in hand and shift+right-clicking, they return a storage component and a regular co-processor. This changes their recipes to match that hard-coded mechanic.

### DIFF
--- a/kubejs/server_scripts/mods/AE2.js
+++ b/kubejs/server_scripts/mods/AE2.js
@@ -917,11 +917,7 @@ ServerEvents.recipes(event => {
     event.remove({ output: /megacells:sky_steel/ })
 
     //MAE2 compat stuff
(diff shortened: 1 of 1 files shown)


ThePansmith:
Not really feeling the skystone bit

PR #933, opened 2024-10-14
Quarter the Trinium requirement for Superconducting Coils

Xefyr0:
This is probably the biggest pain point right before tank. Superconducting coils take 8x the fluid ingredient requirement compared to EBF coils _on the cheapest recipe_, which is unlocked at UV. Moreover, Trinium is hard to come by with the only sources being Naqline and certain Microminer missions which are both somewhat weak.

### DIFF
--- a/kubejs/server_scripts/superconducting_coils.js
+++ b/kubejs/server_scripts/superconducting_coils.js
@@ -1,10 +1,10 @@
 /**
  * A little bit of software wizardry that alters any "superconducting_coil" recipes in the Assembler
- * to use 1/4 of the Trinium they normally do.
+ * to use 1/2 of the Trinium they normally do.
  */
 ServerEvents.recipes(event=> {
     //Get all GTCEu Assembler recipes with an ID matching the regex
-    event.forEachRecipe({ id: /gtceu:assembler\/superconducting_coil/, type: "gtceu:assembler" }, supercon_coil_recipe => {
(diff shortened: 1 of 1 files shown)


Xefyr0:
Regardless of whether you choose to merge this PR or not, I'd like to have a conversation with you about Moni's difficulty over time as players progress - by eliminating or neutering the challenges of Platline, Petrochem and other skill barriers through optional routes (e.g. Platinum from Slime model and Iridium/Osmium through microminers) and AE2's fluid crafting, we are providing a lower level of difficulty compared to base GTM all the way through EV.
This is fine on its own, but if you want to keep the infra check of Trinium intact then players that are used to these "gentle hills" are going to meet with a rude awakening going into LuV when they come across Naqline and the Trinium requirements of Fusion.
With that in mind we should make some conscious decisions about how the difficulty of the modpack ramps up overtime.

Some relevant discussion:
https://discord.com/channels/914926812948234260/1229854271613436066/1298134131112149012
https://discord.com/channels/889188946934923325/1286707174663258228/1297912892129153055

ThePansmith:
Deprecated in favor of #1025,

Xefyr0:
Re-opened this so it's easier to find when the issue Pansmith linked is inevitably resolved if we want to implement it as well.

https://discord.com/channels/914926812948234260/1229929078547550238/1299843089833263177
https://discord.com/channels/914926812948234260/1229929078547550238/1299843587692957777

PR #939, opened 2024-10-14
More resources & ores from GCyR planets

Xefyr0:
This PR is intended to make GCyR planets worth more than a 1-time visit, boost ore veins of materials people struggle to get, and clean up related script files.

- Cleanup and unify GT vein scripts
- Make Fluorite vein also generate in Normal Mode
- Increased yield of Chromite veins in The End
- Added Chromite to Magnetite/Gold veins on Luna
- Added fluid veins to all planets 
- Added equivalent to Moon Dust for all planets, expanded centrifuging products for all variants

I have started work on collection & distillation of planetary atmospheres as well, but the materials they would provide are far more effectively acquired through other means. Let me know if this is something that's desired or if mining & pumping on other planets is sufficient.


Xefyr0:
Yeah I've noticed that people have been struggling with Arsenic - hopefully making it appear more often in its own vein should help.

richie3366:
```and THEN delete it in the next patch after```
@Xefyr0 Does that mean it would also work when someone skips the two updates? Otherwise, I'm afraid it wouldn't happen as rarely as we might think.

File: kubejs/server_scripts/bedrock_fluid.js [hunk 1]
```diff
@@ -1,37 +1,13 @@
...
+
+This is done because pack update mechanisms naively assume files won't be deleted, renamed, or moved;
+They only alter files that are added or modified.
+
+As a result, if this file were not made blank then a player would update their pack to have both this old file
+AND the new file both exist at once with identical contents.
+
+This causes errors in KJS or other difficult-to-troubleshoot issues, 
+so we first update this old file to something nonfunctional as you see here
+and THEN delete it in the next patch after.
```
richie3366: @Xefyr0 Does that mean it would also work when someone skips the two updates? Otherwise, I'm afraid it wouldn't happen as rarely as we might think.

Xefyr0: No, it wouldn't.

It was an attempt to mitigate the problems that arose when people updated the modpack [as per this suggestion](https://discord.com/channels/914926812948234260/1297246927204778077/1297246927204778077), but I don't think I'm going to continue with the idea since:
1. It's a problem with launchers, and trying to fix it from our end will require a lot of extra work
2. Any effort we put in to fix it on our end won't matter if someone makes a mistake in one commit
3. CF launchers do not have this issue (Prism too, as of the 9.0 update)

So I'm thinking the best way forward is to forego this approach and encourage people to switch launchers if they have this issue in the future

PR #941, opened 2024-10-15
Increase Recipe Output For AE2 Processors

three-fold:
Increase the AE2: Logic, Calculation & Engineering Processor recipe outputs from 1 to 2 in the Circuit Assembler only.

Sure, the player can make use of the overclocking features in the Circuit Assembler, but arguably the cost of adding Soldering Alloy and having the Cleanroom requirement on top of that is not worth switching away from the Inscriber.

This change encourages swapping recipes for more efficient methods of producing the Processors and rewards the player for reaching IV.


Echoloquate:
The recipe is fine imo, the main reason for using it is when you get the multiblock that can run in parallel

PR #955, opened 2024-10-16
Add tooltips for Telepastries cakes

### LINKED ISSUE #893: [Bug]: Cakes don't tell you what they're refilled with
### Issue Summary

(Harder mode, moni version 0.8.1)

quest

The quest about teleporting cakes tells you they show what they're refilled with in their tooltips, which is not true.

tooltip

Wasn't sure if it's a mod misconfig, broken mod, or just a lie in the quest description, so submitted as a separate bug

### Reproduction steps

1) Look at the cake tooltip
2) See nothing about how to refill

### Relevant log output

_No response_

ComedyOutput:
Adds destination and refill item to the tooltips for all cakes except Lost City cake, as that one didn't work.
Lost City cake could not refill without also teleporting to destination, which uses the refill instantly.

Closes #893


Xefyr0:
Added a tooltip to the Lost City cake that does as you describe & compacted their tips to one line in 2809facbdb970c6297947636e198b46293d5325e

PR #971, opened 2024-10-19
Nerf solars' energy storage

Xefyr0:
Hello, it's me.
Nerfing solar panels again, but hopefully this time will be the last.

In Monifactory, solar spam is an effective, but boring strategy.
While my previous nerfs helped pull the intensity of spam away from Tier 7 solars specifically by buffing Sculk solars in the process, it didn't address the heart of the issue, that being that Solar power is _extremely_ spammable, and is easily viable as a base's sole source of power generation, especially going into the lategame.

Unfortunately, players will _always_ choose the option with minimum upkeep regardless of cost or output and simply scale it up. A fantastic example of this was Botania's passive generating flowers before they were removed, and even in Moni we see people spamming Naq IIs post-tank instead of trying out the new Cryococcal Resonator.

Thus, this PR's approach is simple: Since players can (and will) always simply place more solars, we take a page from Factorio's book and require the player to invest in power storage infrastructure as well.

Solars don't always output their maximum power, even in the Void dimension where it is always day. When it rains in the Overworld, all solars in the Overworld and JAVD will see their generation rate quickly drop to about 30% (previously 60%) of their maximum. If a player's base consumes more power then the solars can generate at that time, then the solars will draw from their internal energy storage to continue supplying the base. 

Previously, this internal storage was quite generous, capable of easily holding off a player's energy consumption on their own until the sun came out again. With this PR, the solars' storages have been nerfed so that players will _have_ to turn to another energy generation approach at that time, or else have sufficient external power storage to hold out during the storm.

The last thing to note is that this nerf disproportionately affects higher-tier solars, since solar spam is largely an endgame phenomenon when players have the infrastructure to perform deep automated crafts en masse - now those same players have to use that same infrastructure to also make energy storage.


Xefyr0:
As a side note, there's another reason players choose to go for Solar power in LuV+ specifically: Plasma power is _absurdly_ expensive to make for its power output, and Naquadah reactors are gated to Omnium. If there are plans to merge this PR and #933 both at some point, doing them at the same time (or doing 933 first) would make the transition to Plasma significantly easier.

PR #985, opened 2024-10-20
Localized keys

ppoozl:


ThePansmith:
Have you considered looking at [ftb quest localizer](https://www.curseforge.com/minecraft/mc-mods/ftb-quest-localizer)?

ppoozl:
OK. I will take a look at that MOD. Main goal is to do translation. I am using my freetime to do some translate. I have my own fulltime job. I like this pack so I want to localize it so more people can enjoy it.

PR #989, opened 2024-10-20
Post-tank circuit rework

Xefyr0:
Fixes:
- Quantum SoC Cutter recipe having inconsistent output amounts depending on lube choice
- Post-tank supercomputers not being named as such
- Missing cleanroom requirement for circuits in the Circuit Assembler

Refactored:
- GTCEu circuit recipe alteration so that we don't have to write the whole recipe out each time an item is replaced

Reworked:
- Complex SMD recipes
- Post-tank circuit recipe ingredients
- Post-tank circuit recipe voltages & durations

Added:
- SoC recipe for Matter processors using Quantum SoC


ThePansmith:
Merging this so the new function and fixes are present, the recipes might be adjusted again once those new materials are added

PR #993, opened 2024-10-21
Rebalance Functional Storage

Xefyr0:
At this point in time, Functional Storage completely outclasses Super/Quantum Chests for item storage, and is not often used (if at all) for fluids because it requires Super Tank IIIs.

To fix that, this PR makes the following changes:
 1. Upgrades are significantly less effective overall, with each copper upgrade only doubling storage for items instead of _octupling_ it. They are still stackable to exponential effect, with 4 giving a 16x storage multiplier to the base of 2048 items stored for a total of a little over 32k.
 2. Upgrades require GT drums/crates/supertanks/superchests to craft. This means that Functional Storage drawers are cheaper for bulk storage early on, but more storage per-block means getting a super/quantum tank/chest. However, those can then be crafted into storage upgrades for even higher per-block storage with Functional Storage once again.
 3. Fluid drawers only require Iron and Glass.


Xefyr0:
why this sort of thing is needed: https://discord.com/channels/914926812948234260/1229854271613436066/1299242713173983283

richie3366:
> why this sort of thing is needed: https://discord.com/channels/914926812948234260/1229854271613436066/1299242713173983283

But doesn't 32/2 equal 16/1? (`NETHERITE_MULTIPLIER/FLUID_DIVISOR`)

Xefyr0:
> But doesn't 32/2 equal 16/1? (`NETHERITE_MULTIPLIER/FLUID_DIVISOR`)

Yes. Unfortunately, there are two factors of Functional Storage that absolutely wreck a clean integration into GT progression:
1. Functional storage upgrades stack to 4, and their benefits are multiplicative, not additive. There is no config to reduce this.
2. There is no config to alter the base storage of an item or fluid drawer.

Ideally there would be a config to limit the amount of upgrades allowed per drawer to 1 or 2, but since that's not an option (and gating things harder or removing upgrade tiers entirely would be met with a very negative reception) the best option I can think of is to do what's in this PR.

ThePansmith:
The problem with trying to nerf the storage upgrades to such a degree is that it then becomes far easier to just spam low tier drawers then invest in upgrading them.

Xefyr0:
> The problem with trying to nerf storage upgrades to such a degree is that it then becomes far easier to just spam low tier drawers then invest in upgrading them.

The thing is, it's going to take a *lot* of spam to match stacked upgrades - 4 copper upgrades (lowest tier) is equivalent to 16 base drawers.
4 gold upgrades is 256 of the base drawers.
4 diamond upgrades is 4096 base drawers.
4 emerald upgrades is 65536 base drawers.

This is all after the changes in this PR, of course.

ThePansmith:
While possibly true, that really only starts mattering once you get an ME network up and storage stops mattering. In earlygame, storage drawer spam would def be a thing, given the player doesnt really have too many other options anywho.

Xefyr0:
OK.
So how would you suggest balancing Functional Storage over the voltage tiers, then?
Fluid drawers even barely beat out Super Tank X's (UHV) when fully upgraded, an achievement that can be made _pre-machinery._

ThePansmith:
You said it yourself, its impossible to balance, but if i had to pick something, likely just decrease the exponential-ness of the drawers down from 2-4-8-16 to like 2-3-4-5. It would def nerf the drawers, but wouldnt hurt earlygame players that much.

Or just leave it as is cause realistically that shit is not getting balanced well and i would rather have the fun option over the harshly nerfed one lol

Xefyr0:
Done.

desagas:
IMO, any pack that limits storage as you are debating here is not worth playing. Gregtech requires literal orders of materials to complete in full. The limit should not be "what matches the voltage, it should be what exponentially is required to create a functioning machine at EACH teir. IE, MVis 4x that of LV, while HV is 4x or more of that of MV, because of the recursive crafting necessary and the "use lower tiers to craft higher tiers" source of new GT components. Thus, having it so one voltage tier of storage only 2x-ing is ridiculous, and does not in fact follow the regular GT progression in this pack, yet alone in any other pack.

Again, this is just my opinion.

desagas:
This also creates a big issue when interacting with drawers that currently are over the new change limit thresholds, where 100% of the overage will enter the users inventory. Which means I just got a stack of 16972 netherrack by simply right clicking on the drawer to put more into the drawer.

PR #1002, opened 2024-10-22
Sophisticated Storage rework

Xefyr0:
- Both backpacks and storages now follow the same unified GT/TE progression, albeit with different amounts of the same materials
- Storages are now crafted directly from the previous tier like backpacks
- All tiers of backpack and storage are now craftable
- Shulkers are properly obtainable and appear in JEI
- Textures & lang updated to reflect all of the above changes and Gregtech: Refreshed colors
- All recipes show properly in JEI (or at least to the degree that the mod does regularly)
- Compacting upgrades limited to EM only and are now obtainable for other modes
- Replaced moni:gemblock and moni:sophcontainer tags with preexisting tags
- Tier upgrade items and related tooltips are removed & hidden, but still functional for people who made them
- Removed some redundant lines from .gitignore


ThePansmith:
Missing the applicable upgrade items

Xefyr0:
If you're talking about the Actually Additions Storage Crates, they have 117 slots (9x13) while the Iron/Steel (Gold) barrels have 81 slots. (9x9)
The Aluminum (Diamond) barrels would be closer at 108 slots, (9x12) but more importantly why even bother with matching them? SophStorage barrels are not ActuallyAdditions storage crates; they have upgrades and different recipes. If you wanted to match them why were the recipes not matched?

juh9870:
Imo, we don't need all tiers for storage. Having "gold" chest being gated behind aluminium is just an unnecessary early game padding. And gating the top storage tier behind Tungsten is meaningless, because at that point players would just use AE. The current system of Iron (gold) -> Aluminium (diamond) -> Stainless (Netherite) works good as is.

ThePansmith:
looking at this again, I agree with above. 

Though, could swap out iron (gold) for steel, and make the actual iron barrel into bronze for HM+ players.

Xefyr0:
Done. 👍

PR #1014, opened 2024-10-24
Add Ensorcellation + Config

### LINKED ISSUE #1013: [Feature]: Add Ensorcellation for utility enchantments to assist with building
### Suggestion

[Ensorcellation](https://www.curseforge.com/minecraft/mc-mods/ensorcellation) is a mod by Team CoFH that adds various enchantments. Most are related to combat and should probably be disabled in config for this pack, but there are two enchantments that I bet almost every builder would appreciate.

- Reach on chestplate increases block placement/destruction range, which means you don't have to get nearly as close to the ground when flying to work on your factory. Additionally it would further incentivize use of a magnet, so you can actually pick up blocks that you break with your increased reach.

- Air Affinity on helmet removes the penalty to block breaking while not on ground. Red Hook already does this, so having a way to break blocks at a reasonable speed while flying would increase QoL for Angel Ring users (Jetpack users too unless that mod already has a solution for this).

Whether the mod is added to the pack or made into optional compat, these enchantments should get EnderIO enchanter recipes as well.

Echoloquate:
Disabled most enchants because they are combat related, just Air Affinity (removes flying penalty from breaking blocks) and Reach (Exactly what it says)

Closes #1013

### DIFF
--- a/config/NuclearCraft/materials.toml
+++ b/config/NuclearCraft/materials.toml
@@ -265,7 +265,6 @@
 	vein_size = 7
 	min_height = -60
 	max_height = 60
-	veins_in_chunk = 3
 
 [platinum]
 	register = true
@@ -273,7 +272,6 @@
 	vein_size = 7
 	min_height = -60
 	max_height = 0
-	veins_in_chunk = 3
 
 [thorium]
 	register = true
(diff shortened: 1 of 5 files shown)


ThePansmith:
Add recipes for these books in EIO's enchanter

Echoloquate:
Added them, I figured the effortless building reach upgrade was cheap enough to be ok for reach but happy to change it to something else

Echoloquate:
Actually I've got better items, don't merge yet

Echoloquate:
image
image

PR #1015, opened 2024-10-25
Texture update part 2 + Superassembler buff

WithersChat:
- Updated a bunch of block textures to fit GTr
- Superassembler now uses UEV machine parts (much less infinity) and Cryococcus fusion casings (mk IV?)
- Netherite casing is now cheaper since Hyperbolic Microverse Projector uses it


WithersChat:
Oh also, this includes the changes in #991

PR #1042, opened 2024-10-29
Fix enderio redstone conduits for some blocks

richie3366:
While we wait for a fix from EnderIO (see https://github.com/Team-EnderIO/EnderIO/issues/852#issuecomment-2395092479)
javaw_8EjpeG9KLF


Xefyr0 (changes requested):
Minor nitpick: event.add can work with an array in the second parameter to apply the tag to every item in the array, removing the need for the forEach. so for example, you could assign the array to the variable "list" and after the definition simply go
event.add('enderio:redstone_connectable', list)

richie3366:
Thx, I wasn't aware it was possible to provide an array, it's indeed better and most likely more efficient since it reduces JS<->Java transactions

PR #1067, opened 2024-11-06
Update simulating_resources.snbt

hbliOs:
it was statet : "&bHNN&r machines are not sided, so you can input and output from any side."

so there was a mismatch

loot fabricators are sided


ThePansmith:
Already covered in 1003

PR #1077, opened 2024-11-10
Nuclearcraft: Neoteric update

Xefyr0:
Updates Nuclearcraft: Neoteric to the most recent beta, despite the developer's warning.
Also vastly increases fission fuels' FE output and lifetime now that it's possible, gates higher-tier nuclear fuels behind thermal centrifuge recipes, and rebalances Sunnarium to avoid conflicts with those tiered thermal centrifuge recipes.

### DIFF
--- a/config/NuclearCraft/fission.toml
+++ b/config/NuclearCraft/fission.toml
@@ -1,69 +1,14 @@
 
 #Settings for reactor fuel
 [reactor_fuel]
-	#Heat multiplier for boiling reactor.
+	#Heat multiplier affects on heat/cooling ratio multiplier.
 	heat_multiplier = 3.24444444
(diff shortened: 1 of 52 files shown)


config/NuclearCraft/fission.toml
liukebupt, line 1: Wrong value changed

config/NuclearCraft/fission.toml
```diff
@@ -7,7 +7,7 @@
 	#Range: 0.0 ~ 1000.0
 	depletion_multiplier = 100.0
 	#Heat multiplier. Affects to all fuels.
-	#Range: 0.01 ~ 100.0
+	#Range: 0.01 ~ 300.0
```
liukebupt, line 10: Wrong value changed

PR #1120, opened 2024-11-15
Restore alternative reciepe for oak drawers using only chests.

### LINKED ISSUE #782: [Bug]: Functional storage drawer recipe isn't working correctly
### Issue Summary

When crafting drawers from functional storage the amount of wood you are using doesn't matter only the amount of chests in the middle, example:

64 chests surrounded by 8 wooden planks, the output should be 1 drawer.
image

64 drawers are crafted when shift clicking to craft
image

### Reproduction steps

1. Open handheld crafting table or crafting table
2. Put larger amount of chests in the middle slot of the recipe and 1 wood in all other slots
3. Shift click on the output of the craft and you will get the amount of drawers that you put in chests

### Relevant log output

_No response_

tomprince:
These were removed in #826 to fix #782. However, I think the behaviour is working (at least mostly) as intended. It is definitely intended that you can craft oak drawers without needing planks, shift clicking allowing crafting different recipes with the same output is slightly surprising.

This is a slightly cheaper, but avoids there being a micro-optimization of using 4x drawers instead of 1x drawers as crafting ingredients (since 4x drawers cost 9.25 planks vs 16 for 1x drawers). With these recipes, they can both cost 8 planks unless you want different textures.

### DIFF
--- a/kubejs/server_scripts/mods/functionalstorage.js
+++ b/kubejs/server_scripts/mods/functionalstorage.js
@@ -1,8 +1,4 @@
 ServerEvents.recipes(event => {
-    //Remove bugged Oak Drawer recipes
-    event.remove({ id: /functionalstorage:oak_drawer_alternate/ })
-
-
     // Fluid Drawers
     event.replaceInput({ output: /fluid/, mod: 'functionalstorage'}, '#minecraft:planks', '#forge:plates/iron')
     event.replaceInput({ output: /fluid/, mod: 'functionalstorage'}, 'minecraft:bucket', 'gtceu:lv_hermetic_casing')


Xefyr0:
Unfortunately, the behavior is most certainly _not_ working as intended.
If you decompile the mod jar and look at the recipe JSONs, you can see that the original recipe expects any item with the `functionalstorage:woodless` "type", which is as far as I know is invalid and thus ignored. This causes the recipe to complete without any such item necessary.

The micro-optimization argument also falls flat, since there are similar bugged recipes for the 1x2 and 2x2 oak drawers.

If you don't feel like decompiling a mod jar, you can find the recipe JSON files attached below:
[oak_drawer_alternate_x1.json](https://github.com/user-attachments/files/17769695/oak_drawer_alternate_x1.json)
[oak_drawer_alternate_x2.json](https://github.com/user-attachments/files/17769697/oak_drawer_alternate_x2.json)
[oak_drawer_alternate_x4.json](https://github.com/user-attachments/files/17769698/oak_drawer_alternate_x4.json)

Frankly, if you want to go report this bug on Functional Storage's repo on our behalf that'd be very helpful:
https://github.com/Buuz135/FunctionalStorage/issues

tomprince:
> The micro-optimization argument also falls flat, since there are similar bugged recipes for the 1x2 and 2x2 oak drawers.

I'm not sure what you mean. My point is that the default recipies for the 1x1, 1x2 and 2x2 drawers take 16, 11.5 and 9.5 planks respectively. So, if you care about optimizing plank usage, it makes more sense to make 2x2 drawers as input to crafting recipies (such drawer upgrades, compacting drawers, controllers). The alternative recipes cost a flat 8 planks per drawer.

Particularly when using framed drawers, it seems sensible to allow you to craft drawers without needing a specific type of plank.

tomprince:
Also, whether or not the recipes are intentional, I find them a nice QoL things that lets you not worry about what type of wood you have, but get a consistent style of drawers.

PR #1129, opened 2024-11-15
Assorted circuit tweaks

Xefyr0:
- Make Post-tank processors cheaper after accidentally making them more expensive
- Fix Dimensional circuit board, processor texture being one pixel too high
- Reduce inflated Niobium-titanium cost of Wetware
- fix HM wetware requiring liquid tin instead of Sterilized Growth Medium


kubejs/server_scripts/gregtech/circuits.js
```diff
@@ -145,7 +172,7 @@ ServerEvents.recipes(event => {
     // WIP: Dimensional Circuits
     event.recipes.gtceu.circuit_assembler('dimensional_circuit_board')
         .itemInputs(
-            '32x kubejs:matter_circuit_board',
```
cewlboi, line 148: qb says matter circuit boards will be used to craft remaining boards as well; this would change that

Xefyr0: Fixed in latest commit

Xefyr0:
Git confuses me.
Quests are fixed, though

PR #1137, opened 2024-11-18
Add a Programmed Circuit to the Ender Chest assembler recipe to avoid conflicts with the Flux Core

StarskyXIII:
When bulk crafting, the machine may occasionally prioritize using the ingredients to craft Ender Chests.

image
image


StarskyXIII:
Due to the recipe of the Ender Chest not requiring a Programmed Circuit, 
even if the Input Bus has a Programmed Circuit set, 
when batch crafting Flux Core, it will still prioritize crafting the Ender Chest.

PR #1141, opened 2024-11-21
Optional mod compatibility: Advanced peripherals

Niki4tap:
[Advanced peripherals](https://github.com/IntelligenceModding/AdvancedPeripherals) is a mod that expands the functionality of an already included optional computercraft, it allows computers to send chat messages, detect player presence, and much more.

This patch makes all recipes gregified and compatible with monifactory.

### DIFF
--- a/kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
+++ b/kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
@@ -1,0 +1,142 @@
+const modid = "advancedperipherals"
+
+if (Platform.isLoaded(modid)) {
+    if (!Platform.isLoaded("computercraft")) {
+        console.log("Advanced peripherals is installed, but computercraft is not! Not running compat scripts for advanced peripherals.")
+        return
+    }
+
+    console.log("Advanced peripherals found, loading compat scripts...")
+    ServerEvents.recipes(event => {
+        const replaceRecipe = (output, shape, dictionary) => {
+            const id = `${modid}:${output}`
+            event.remove({ id: id })
+            event.shaped(id, shape, dictionary)
+        }
+
+        replaceRecipe("peripheral_casing", [
+            'CMC',
+            'SLS',
+            'CMC'
+        ], {
+            C: "computercraft:cable",
+            M: "computercraft:wired_modem",
+            S: "#gtceu:circuits/mv",
+            L: "gtceu:mv_machine_hull"
+        })
+
+        replaceRecipe("chat_box", [
+            'BEB',
+            'PCP',
+            'ASA'
+        ], {
+            B: "minecraft:book",
+            E: "gtceu:mv_emitter",
+            P: "minecraft:ender_pearl",
+            C: `${modid}:peripheral_casing`,
+            A: "minecraft:paper",
+            S: "gtceu:mv_sensor"
+        })
+
+        replaceRecipe("energy_detector", [
+            'C C',
+            'TPT',
+            'C C'
+        ], {
+            C: "gtceu:copper_single_cable",
+            T: "gtceu:mv_transformer_1a",
+            P: `${modid}:peripheral_casing`
+        })
+
+        event.remove({ id: `${modid}:environment_detector` })
+        const envDetectorRecipe = (plate) => {
+            event.shaped(`${modid}:environment_detector`, [
+                'PSP',
+                'SCS',
+                'DSD'
+            ], {
+                P: plate,
+                S: "gtceu:lv_sensor",
+                C: `${modid}:peripheral_casing`,
+                D: "minecraft:dirt"
+            })
+        }
+        envDetectorRecipe("gtceu:nether_quartz_plate")
+        envDetectorRecipe("gtceu:certus_quartz_plate")
+        envDetectorRecipe("gtceu:quartzite_plate")
+
+        replaceRecipe("player_detector", [
+            'OSO',
+            'PCP',
+            'MSM'
+        ], {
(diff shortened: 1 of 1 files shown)


kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
```diff
@@ -0,0 +1,142 @@
...
+            'ETE',
+            'RPR',
+            'ECE'
+        ], {
+            E: "minecraft:ender_pearl",
+            T: "minecraft:crafting_table",
+            R: "gtceu:mv_robot_arm",
+            P: `${modid}:peripheral_casing`,
+            C: "minecraft:chest"
+        })
```
tomprince: This has infinite range, so should probably be higher tier. Also, have some of emiiter/sensor/field generator.

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
```diff
@@ -0,0 +1,142 @@
...
+            'SCS',
+            'GTG'
+        ], {
+            G: "ae2:fluix_glass_cable",
+            I: "ae2:interface",
+            S: "#gtceu:circuits/mv",
+            C: `${modid}:peripheral_casing`,
+            T: "ae2:pattern_provider"
+        })
+    })
```
tomprince, line 140: The ME bridge allows you to implement ME requester functionality with computercraft. That suggests that it should be further in the progression than MV. I do think that it might be reasonable to use an earlier tier than the requesters though, perhaps EV?

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
```diff
@@ -0,0 +1,142 @@
...
+        envDetectorRecipe("gtceu:certus_quartz_plate")
+        envDetectorRecipe("gtceu:quartzite_plate")
+
+        replaceRecipe("player_detector", [
+            'OEO',
+            'PCP',
+            'MSM'
+        ], {
+            O: "minecraft:observer",
+            E: "gtceu:mv_emitter",
```
tomprince: Since this block is input only, it probably doesn't need an emitter.

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
Niki4tap: > That suggests that it should be further in the progression than MV.

I would disagree, export bus with a craft card on a drawer makes a great requester, which me and my friend used to use in nomifactory (worked very well there) and are using here because ME requesters are buggy (putting them behind a toggle bus makes them refuse to work, cannot input fractional number into fluid request), so if anything the requesters should go down in tier (they're fine where they are for their "convenience" though (if they worked)).

> perhaps EV

Another point here would be that we used and are using the ME bridge for very simple functionality like checking how much energy a network consumes, which, again, imho, should not require EV components.

And in the end of the day, ME requester functionality isn't something game breaking, and considering that implementing it in cc would require quite a bit of effort, I would say MV is a nice tier for it.

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
Niki4tap: Sure, do you think another sensor instead would be fine?

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
Niki4tap: Honestly, haven't used this peripheral, but yeah seems pretty powerful, I would maybe gate it behind EV? (I think that's where ae2 wireless terminals start)

Does this recipe look good?
```js
[
    'WEW',
    'RPR',
    'FCF'
], {
    W: "ae2:wireless_receiver",
    E: "minecraft:ender_pearl",
    R: "gtceu:ev_robot_arm",
    P: `${modid}:peripheral_casing`,
    F: "gtceu:hv_field_generator",
    C: "minecraft:chest"
})
```

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
tomprince: Yeah.

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
tomprince: That looks reasonable, though I think I might go with a HV robot arm. (The wireless receivers require an EV emitters, providing the EV gate).

kubejs/server_scripts/mods/optionalCompats/advanced_peripherals.js
tomprince: > And in the end of the day, ME requester functionality isn't something game breaking, and considering that implementing it in cc would require quite a bit of effort, I would say MV is a nice tier for it.

While it may not be game breaking, the requester was recently pushed back from HV to LuV. It seems like providing equivalent functionality at MV in a compat mod doesn't make sense. I do agree that it requires a bit of effort to implement using CC (though I think you can find scripts online that save a bunch of that effort), which is why I suggested EV (which is two tier before you can get ME requesters).

(Note than I am not one of the developers of this pack, so can't make the final call.)

Niki4tap: Wow. LuV for requesters? Yeah, no, I think I'll take my chances, leave as is, and see what main devs say here.

(Thanks for the comments still)

PR #1161, opened 2024-11-25
Support monicoins in compacting drawers in hard/expert.

tomprince:
Also, remove the unused monipenny in those modes.

### DIFF
--- a/kubejs/server_scripts/monicoins/monicoins.js
+++ b/kubejs/server_scripts/monicoins/monicoins.js
@@ -1,13 +1,15 @@
 ServerEvents.recipes(event => {
-    event.shapeless('5x kubejs:moni_penny', 'kubejs:moni_nickel')
-    event.shaped('kubejs:moni_nickel', [
-            ' A ',
-            'AAA',
-            ' A '
-        ], {
-            A: 'kubejs:moni_penny'
-        }
-    )
+    if (isNormalMode) {
+        event.shapeless('5x kubejs:moni_penny', 'kubejs:moni_nickel')
+        event.shaped('kubejs:moni_nickel', [
+                ' A ',
+                'AAA',
+                ' A '
+            ], {
+                A: 'kubejs:moni_penny'
+            }
+        )
+    }
 
     event.shapeless('5x kubejs:moni_nickel', 'kubejs:moni_quarter')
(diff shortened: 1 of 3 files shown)


kubejs/startup_scripts/registry/item_registry.js
```diff
@@ -458,3 +460,12 @@ StartupEvents.registry('item', event => {
...
 })
+
+ItemEvents.toolTierRegistry(event => {
+    event.add("dev", (tier) => {
+        tier.enchantmentValue = 42
+        tier.level = 42
+        tier.uses = 42690
+        tier.speed = 12
+    })
+})
```
Niki4tap: Unrelated change?

kubejs/server_scripts/monicoins/monicoins.js
```diff
@@ -1,13 +1,14 @@
-ServerEvents.recipes(event => {
```
Niki4tap, line 1: Pretty sure this line shouldn't be removed

OrioPrisco: i think so too, but if you really wanted to remove it properly you'd need to remove the matching parenthesis at the end of the file

kubejs/startup_scripts/registry/item_registry.js
tomprince: It looks like both of these are artifacts of a bad rebase. Fixed.

Niki4tap: Ah, happens, it looks like the other comment is still valid though...

ThePansmith:
All good, but removing monipenny is kinda random, keep it

tomprince:
> All good, but removing monipenny is kinda random, keep it

I wanted to *not* add monipenny -> moninickel compacting recipe in hard/expert, to avoid getting a penny-nickel-quarter drawer if you put the wrong item in a drawer first. There is no way (other than downcrafting) to get any pennies, since they are not given as quest rewards. To be consistent with the compacting drawers, I thought it makes sense to remove the coin, given that there are no uses in hard/expert. 

 since none of the quests give you them, and the only way to get them would be to manually craft them (for no useful purpose).

ThePansmith:
actually yeah fair enough

PR #1165, opened 2024-11-25
Fix whitespace.

tomprince:
This fixes a bunch of whitespace, particularly around all the `if (is*Mode)` checks.


tomprince:
You can see a diff with the whitespace changes ignored [here](https://github.com/ThePansmith/Monifactory/pull/1165/files?diff=split&w=1).

PR #1168, opened 2024-11-25
Update gregtech

tomprince:
I've not (yet) tested this, but I thought I'd push this PR to save somebody else needing to look up the file ID for the manifest.

(Also includes #1167 so that downloading mods works)


tomprince:
I also update the config files by starting and quitting the game (plus incorporating #1162 into the config-overrides).

PR #1188, opened 2024-11-30
Tweak Various Microminer Recipes to push back some line skips, buffed Fusion MK3

Xefyr0:
There have been plenty of discussions about this, but some persons have been concerned with the ease in which players can skip certain processing lines, citing the argument that by avoiding these challenges, they arrive at the endgame ill-prepared for the difficulty spike it presents.

This PR offers a solutions for some of these, each separated into an individual commit for easy cherry-picking:

1. Discourage players from skipping Platline by removing the T4MM mission that returns Osmiridium.
Without this mission, all ores that yield Osmium and Iridium in reasonable amounts are gated behind T5MM (itself made of Iridium) or Platline.
Note that this leaves Chalcocite/Cuprorhodsite (Ruthenium ores) as-is.

2. Discourage players from skipping Naqline by removing the mixer recipes for Enriched Naquadah/Naquadria.
Without this mission, there is no other source of Enriched Naquadah/Naquadria before doing Naqline.
Note that this leaves Kaemanite (Trinium ore) as-is.

3. Discourage players from skipping Mk3 fusion (NM only) by reverting the buff for the T9MM mission that returns Neutronium.
I have read messages from a group of players who cited Mk3 fusion to be useless given the presence of this mission. Given that Mk3 is so close to Creative Tank _and is otherwise useless due to Tank,_ then this may be intended behaviour so as to be rid of what may amount to an annoying obstacle. Otherwise, this would make players use Mk3 fusion once more.

### DIFF
--- a/config/ftbquests/quests/chapters/late_game.snbt
+++ b/config/ftbquests/quests/chapters/late_game.snbt
@@ -789,7 +789,7 @@
 				""
 				"You can also get tons of this from &6Platinum Group Sludge&r."
 				""
-				"&2Osmiridium 80/20 Ore and Iridosmine 80/20 Ore&r is available renewably from the Tier Four and Tier Six micro miners, as well as semi-renewably from the Tier Five micro miner."
+				"&2Osmiridium 80/20 Ore and Iridosmine 80/20 Ore&r are available renewably from the Tier Six micro miner, as well as semi-renewably from the Tier Five micro miner."
 			]
 			id: "025092551C0A20B6"
 			rewards: [{
--- a/config/ftbquests/quests/chapters/late_game.snbt
+++ b/config/ftbquests/quests/chapters/late_game.snbt
@@ -179,14 +179,6 @@
 			y: 8.25d
 		}
 		{
-			id: "26C607D68B1B1404"
-			linked_quest: "1C1B1E1BBFA079F5"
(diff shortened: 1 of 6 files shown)


Xefyr0:
Adjusted quests for platline/naqline skip removals & rebased onto main.

If we do decide to remove platline/naqline skips, make sure to merge both relevant commits for each.

ThePansmith:
After putting it through the form and hearing what people have to say, here are my thoughts:

1. People agreed with this, though one did mention the possible issue of T4 possibly becoming useless as a result of it, which I do agree with. Perhaps replace the 80/20 mission with something related to platline; maybe bulk amounts of Sheldonite (most viable PGS method, currently has no MM mission) or PGS itself? Could throw in salt as a secondary reward as well.

2. Yeah just pull this one

3. General consensus is that MK3 is just not really good, it kinda has the same issue as with bromine where it exists for all of like 2 hours before it gets replaced by tank/stabminers. Leave the MM mission alone, but also buff the MK3 recipe. The NT MK3 recipe is by far and away the longest fusion recipe. Could lower the recipe down to 6.5 seconds per (which would allow it to still keep that title, but still make the MK3 35% faster) or multiply the amount of material it processes in a given recipe, etc.

PR #1200, opened 2024-12-03
Recipe ideas

Xefyr0:
A set of suggestions for recipe changes/additions.

Each atomic change is in its own commit like the Platline/Naqline/Fusion Mk3 skip removal PR.

Some of these are more strongly suggested than the others. (namely the commits whose names start with "Fix")

### DIFF
--- a/kubejs/server_scripts/gregtech/Alloys_Recipes.js
+++ b/kubejs/server_scripts/gregtech/Alloys_Recipes.js
@@ -120,13 +120,17 @@ ServerEvents.recipes(event => {
         .duration(140)
         .EUt(30)
 
-    //Replace default GTCEu glowstone separation recipe to match mixing recipe
-    event.replaceOutput({ id: "gtceu:centrifuge/glowstone_separation" }, 'minecraft:redstone', 'gtceu:tricalcium_phosphate_dust')
-    event.recipes.gtceu.mixer("kubejs:glowstone_dust")
-        .itemInputs('gtceu:tricalcium_phosphate_dust', '#forge:dusts/gold')
-        .itemOutputs('2x minecraft:glowstone_dust')
+    //Replace default GTCEu glowstone mixing/separation recipes
+    event.recipes.gtceu.mixer("gtceu:mixer/glowstone_mixing")
+        .itemInputs('gtceu:tricalcium_phosphate_dust', 'gtceu:barite_dust', '#forge:dusts/gold')
+        .itemOutputs('3x minecraft:glowstone_dust')
         .duration(80)
         .EUt(15)
+    event.recipes.gtceu.centrifuge("gtceu:centrifuge/glowstone_separation")
+        .itemInputs('3x minecraft:glowstone_dust')
+        .itemOutputs('gtceu:tricalcium_phosphate_dust', 'gtceu:barite_dust', '#forge:dusts/gold')
+        .duration(976)
+        .EUt(80)
 
     event.recipes.gtceu.mixer("kubejs:mana_infused_dust")
         .itemInputs('kubejs:primal_mana', '#forge:dusts/titanium')
--- a/kubejs/server_scripts/gregtech/reconstruction.js
+++ b/kubejs/server_scripts/gregtech/reconstruction.js
@@ -107,4 +107,36 @@ ServerEvents.recipes(event => {
             .duration(80)
             .EUt(56)
     })
+
+    // Flower cycle convenience recipes
+    const flowerCycle = [
+        'dandelion',
+        'poppy',
+        'blue_orchid',
+        'allium',
+        'azure_bluet',
+        'red_tulip',
+        'orange_tulip',
+        'white_tulip',
+        'pink_tulip',
+        'oxeye_daisy',
+        'cornflower',
+        'lily_of_the_valley',
+        //'sunflower',  //Flowers that give 2x as much dye are excluded. Sorry!
+        //'lilac',
+        //'rose_bush',
+        //'peony',
+        'spore_blossom',
+        'wither_rose',
+        'dead_bush'
+    ] 
+    flowerCycle.forEach((flower, index) => {
+        let curFlower = flowerCycle[index];
+        let nextFlower = flowerCycle[(index+1)%flowerCycle.length];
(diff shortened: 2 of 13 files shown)


kubejs/server_scripts/gregtech/reconstruction.js
```diff
@@ -107,4 +107,36 @@ ServerEvents.recipes(event => {
...
+        'pink_tulip',
+        'oxeye_daisy',
+        'cornflower',
+        'lily_of_the_valley',
+        //'sunflower',  //Flowers that give 2x as much dye are excluded. Sorry!
+        //'lilac',
+        //'rose_bush',
+        //'peony',
+        'spore_blossom',
+        'wither_rose',
```
tomprince, line 130: This already has a reconstructor recipe (dried kelp). And kelp was available from plant balls (probably accidentally removed in #1174). So it might make sense to exclude it.

Additionally, it would probably make sense to add cactus to this list.

kubejs/server_scripts/mods/AE2.js
```diff
@@ -460,7 +460,31 @@ ServerEvents.recipes(event => {
...
+                "middle": {
+                    "item": (wafer[0] + wafer[1])
+                },
+                "top": {
+                    "item": "ae2:silicon_press"
+                }
+            },
+            "mode": "inscribe",
+            "result": {
+                "item": "ae2:printed_silicon"
```
tomprince: I would have guessed that you'd get an increasing number of printed silicon, based on how the different tiers of wafers work in the engraver.

kubejs/server_scripts/gregtech/reconstruction.js
Xefyr0: Maybe.
It might also just make more sense to scrap the flower cycle and have NM players just buy flowers with coins like they can with Cocoa Beans and Cactus - HM/Expert players can do the exploring.

@ThePansmith what do you think?

kubejs/server_scripts/gregtech/Alloys_Recipes.js
```diff
@@ -120,13 +120,17 @@ ServerEvents.recipes(event => {
...
 
-    //Replace default GTCEu glowstone separation recipe to match mixing recipe
-    event.replaceOutput({ id: "gtceu:centrifuge/glowstone_separation" }, 'minecraft:redstone', 'gtceu:tricalcium_phosphate_dust')
-    event.recipes.gtceu.mixer("kubejs:glowstone_dust")
-        .itemInputs('gtceu:tricalcium_phosphate_dust', '#forge:dusts/gold')
-        .itemOutputs('2x minecraft:glowstone_dust')
+    //Replace default GTCEu glowstone mixing/separation recipes
+    event.recipes.gtceu.mixer("gtceu:mixer/glowstone_mixing")
+        .itemInputs('gtceu:tricalcium_phosphate_dust', 'gtceu:barite_dust', '#forge:dusts/gold')
+        .itemOutputs('3x minecraft:glowstone_dust')
```
tomprince, line 126: I'm not sure the motivation behind this change, but this locks synthesizing glowstone to post-nether. It also means that barite is no longer something that you can obviously immediately electrolyze in an ore processing setup.

kubejs/server_scripts/mods/AE2.js
Xefyr0: Good point. Didn't think to look into seeing if AE2 Inscribers can output more than one item at once, but they can.

kubejs/server_scripts/gregtech/Alloys_Recipes.js
Xefyr0: In NM, synthesizing Glowstone is a non-issue thanks to the Witch model providing effectively infinite amounts. A visit to the Nether isn't even necessary to get a Witch model running, actually, since you can buy the ore with Monicoins.
In HM/EM, a visit to the Nether is required pre-LV so it's a barrier for all of 30 mins.

Not sure what you're talking about in regards to the ore processing setup, but I do want to point out that electrolyzing Barite results in Barium, Oxygen, and Sulfur. Sulfur and Oxygen can be found in plenty of other places, and Barium is only useful for GT Superconductors (Most players neglect these for EIO/TE supercoductors) and Omnium (You need to make 10-ish stacks & that's it for all modes)

tomprince: > Not sure what you're talking about in regards to the ore processing setup,

There are a bunch of dusts that you get from ore processing where the *only* thing you can do with them is electrolyse (or centrifuge) them (barite is one of them). I'm inclined to immediately process those dusts in my ore processing subnet. By adding an alternate use, one needs to consider whether to keep it around in the unprocessed form.

kubejs/server_scripts/gregtech/reconstruction.js
ThePansmith: Would be a lot of recipe patterns we would need to make sure don't overlap with preexisting ones. Given that its just dyes, its fine to be an AR recipe.

kubejs/server_scripts/gregtech/Alloys_Recipes.js
Xefyr0: OK, but I don't understand how this would be a bad thing.
Many processing lines start by diverting outputs from oreproc already (Rutile, Bauxite, Uraninite, and more in HM/EM) Heck, Platline requires you to divert Purified Ore, which is in the middle of oreproc instead of the output.

If you're more concerned about not being able to synthesize Barite from Barium/Oxygen/Sulfur despite the singular other source of Barium being Naqline, then that can be solved with a Chemical Reactor recipe.

tomprince: > Many processing lines start by diverting outputs from oreproc already

Sure, but for all of those, the question of "Do I need to divert this?" has the obvious answer of "Yes of course, that is the only (or main) way of getting whatever the product of that processing line". The issue isn't the need to divert something, but rather the difficulty/uncertainty of figuring it out.

It would be one thing if this change improved the gameplay in some fashion, but it seems like it is just to be more "accurate" in describing a fantasy material.

ThePansmith:
Cherrypicked into main

PR #1204, opened 2024-12-04
Allow more tiers of turbo charger in quest.

tomprince:
This allows any tier up to EV (which lapotron crystals) for the quest.

In my most recent run, the first charger I made was HV.

### DIFF
--- a/config-overrides/expert/ftbquests/quests/chapters/genesis.snbt
+++ b/config-overrides/expert/ftbquests/quests/chapters/genesis.snbt
@@ -1833,6 +1833,18 @@
 							}
 							{
 								Count: 1b
+								id: "gtceu:mv_charger_4x"
+							}
+							{
+								Count: 1b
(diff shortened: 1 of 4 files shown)


Xefyr0 (changes requested):
Looks like you forgot about the config overrides.

Also, just a nitpick if you don't mind, maybe move the MV/HV/EV chargers directly below the LV one instead of below the TE charger.

tomprince:
> Looks like you forgot about the config overrides.

Done; though I just applied the patch to the files, and haven't actually tested the chapters.

> Also, just a nitpick if you don't mind, maybe move the MV/HV/EV chargers directly below the LV one instead of below the TE charger.

My intention was that the higher tier charger get displayed last, but in testing, it appears the order does matter. Thus, I've adjusted it as you suggested.

Xefyr0 (approved):

PR #1222, opened 2024-12-06
Fix line-endings in "Into The Microverse" chapter.

tomprince:
Every time I start the game, this file gets rewritten with different whitespace for me. This commits that change, to avoid that in the future.


tomprince:
You can see the (lack) of changes without whitespace [here](https://github.com/ThePansmith/Monifactory/pull/1222/files?w=1).

PR #1224, opened 2024-12-06
Changed functional storage upgrades rarity

MtcLeo05:
I made a mod to remove the hardcoded colors from the functional storage upgrades for a personal modpack and, since I noticed a similar problem in this project, I decided to make a PR where I fix it.

I could not understand how to include the mod to the manifest, so it's not included, but the link is this [Curseforge](https://www.curseforge.com/minecraft/mc-mods/remove-hardcoded-colors-from-functional-storage)

While this mod was not made exactly to work with Monifactory, I have tested it and it did not give problems.

To give an example on how the items would change, here are some screenshots with the before and after
(The colors were color-picked from the textures)
before
after


ThePansmith:
Adding a mod to the manifest is done by adding a new entry, like so.
```
      "fileID": 5430253,
      "projectID": 949555,
      "required": true
```
The fileID is the ID of the specific mod file, easiest way to grab it is via the mod's download url on CF.
ProjectID is displayed in the "about project" section of the mod on CF.

MtcLeo05:
Now the mod should be included in the manifest

PR #1244, opened 2024-12-11
Nuke Perfect Gems

Xefyr0:
Their only uses are to:
- Increase gem yield from microminer missions (no longer necessary with Crystal Matrix rework)
- Intermediate step for Black Steel (fluff)
so this PR gets rid of them seeing as they are essentially bloat.


WithersChat:
Sounds good, I approve.

PR #1269, opened 2024-12-19
WIP GTM 1.6 updates

tomprince:
Posting early to hopeful avoid duplicate work.


tomprince:
This seems to work, with the exception of steamadditions not being compatible with 1.6.

ThePansmith:
Whoops, will wait for GTm 1.6.1 before merging.

kubejs/server_scripts/fixes_tweaks/unification/other_tags.js
```diff
@@ -3,6 +3,4 @@ ServerEvents.tags('item', event => {
 })
 
 ServerEvents.tags('block', event => {
-    event.add('gtceu:cleanroom_floors', 'enderio:travel_anchor');
-    event.add('gtceu:cleanroom_floors', '#elevatorid:elevators');
```
tomprince, line 7: I ended up adding these upstream, so we don't need to add them ourselves.

PR #1301, opened 2024-12-24
Simplify GT crafting component definitions.

tomprince:
I had not looked at any of the high tier recipes, so I couldn't tell if this is apply all the changes appropriately. But it does show how to address the comment at the top of the file.

(cc: @Xefyr0)


tomprince:
Although, thinking about this more I think the interface can be improved on the GT end, to avoid the need for `Map.of`.

tomprince:
See https://github.com/GregTechCEu/GregTech-Modern/pull/2639 for the possible improvement on the GT side.

PR #1303, opened 2024-12-25
Add japanese translation to KubeJS

twister716:


ThePansmith:
What's also nice that weblate will also auto-generate applicable lang files for everything else as well.

PR #1344, opened 2025-01-04
Insane Combustion Engine

Xefyr0:
PR for a Large Combustion Engine above the ECE.
Intake casings are animated, and the multiblock operates as "ECE but bigger" would imply.

Mostly made of Naquadah like how LCE and ECEs are almost entirely Titanium & Tungstensteel respectively, placing the generator at mid-late LuV. However, there is a bit of Enriched Naquadah used in the frames for the casings to gate this almost to ZPM for HM/EM.

This PR also includes a nerf to JEAN, otherwise this addition of a higher-tier combustion engine would be OP for LuV power generation. Roughly 5x as much is needed to produce the same amount of EU.


Xefyr0:
Note for future self: Naquadah and Enriched Naquadah costs should be roughly the same - also ensure that regular Naquadah rotors are used to match the fact that there are only regular Naquadah Turbine Rotors.

WithersChat:
Why was that never merged again?

Xefyr0:
Pansmith did not think this to be necessary.
If fission finally becomes a viable way to produce power, that may be true.
However, I believe it's best to have multiple options for generating power at each stage of the game, and in LuV your only options would be fission and fusion. Players rarely choose fusion for power in LuV due to the infra cost, however, even now when the only other option is scaling combustion generators until Naq Reactors become available.

ThePansmith:
Closing this, with fission changes incoming, it should finally be a viable powergen option

PR #1349, opened 2025-01-05
Fix quartz plate.

tomprince:


Xefyr0:
It seems that there isn't a recipe conflict in dev for some reason - supplying Lubricant nets Quartz Slabs, but Water and Distilled Water nets Quartz Plates. Do you observe this as well, or is there still a conflict on your end?

tomprince:
> It seems that there isn't a recipe conflict in dev for some reason - supplying Lubricant nets Quartz Slabs, but Water and Distilled Water nets Quartz Plates. Do you observe this as well, or is there still a conflict on your end?

I see that, but I'd still consider it an effective conflict. I have a large cutting machine which I supply with lubricant, and have patterns for plates that don't involve fluid. I suspect many players have similar setups.

PR #1351, opened 2025-01-05
Show thermal mob drops in JEI.

tomprince:
This also restores the crafting table recipe for the corresponding powders (which is half as efficient as the macerator recipe).

### DIFF
--- a/kubejs/startup_scripts/_initial.js
+++ b/kubejs/startup_scripts/_initial.js
@@ -59,6 +59,9 @@ global.UnificationExcludedItems = [
     'enderio:energized_gear',
     'enderio:vibrant_gear',
     'enderio:dark_bimetal_gear',
+    'thermal:basalz_rod',
+    'thermal:blitz_rod',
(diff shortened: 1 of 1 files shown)


tomprince:
Actually, I'm not sure if it adds that recipe back, or it was already there. (And my computer is too slow at starting monifactory for me to check)

Xefyr0:
I can see the mob drops regularly on both NM and HM?
blizz
The same is the case for Basalz and Blitz btw.

Moreover, the crafting recipes work normally, too.
What is the benefit of this PR?

tomprince:
But can you see it as an item on the right when you search for it in EMI?

Xefyr0:
Yeah I can, actually.
see

Are you _sure_ there's a benefit to this PR?

Xefyr0:
Closing this PR on the basis of the issue being specific to Tomprince's instance as per [this](https://discord.com/channels/914926812948234260/1229929078547550238/1327834928502669324) message.

PR #1374, opened 2025-01-09
NC:N KJS full replacement

EnsignEvident:
This is still heavily a work in progress. Any advice would be highly appreciated.


ThePansmith:
When making a pull request, the new content should be added to existing files when applicable. For example, the multiblock registery info should go in multiblock_registry.js


Also, the fact that the commits said that you added the files via upload me think you are  manually uploading the files to your github fork. Consider taking a look at the contributing.md file for a better option.

PR #1379, opened 2025-01-09
Tweaks n stuff 2: electric boogaloo

WithersChat:
For context, I made that PR before 0.11 and had to reopen it since I went on hiatus at the wrong time 💀


ThePansmith:
May be best to cherrypick your changes into a fresh branch

PR #1382, opened 2025-01-10
Exoskeleton Legs

Xefyr0:
This mod adds Exoskeleton Legs, which are intended to be a replacement for Travellers' Boots.
Equip them and you get a speed boost + full-block step height.

Pros:
- No config file getting modified every goddamn time you launch the game
- No particle effects
- No durability
- Frees up Boots slot for other armor sets

Cons:
- No jump boost

Also trying to deprecate items before removal for the first time - remember to remove the mod & relevant KJS script in 0.12.0!


ThePansmith:
Standardized some things in #1383, template literals are used instead of string concatenation.

PR #1401, opened 2025-01-13
Update taranium.js

Jollahs:
Makes xenic acid breakdown stoichiometrically correct


Jollahs:
the removal of hydrogen peroxide from the recipe is consistent with other acid electrolysis where oxygen gas is a byproduct, not H2O2

PR #1404, opened 2025-01-13
chore(localization): add Chinese translation for "GTCEu Native"

marioplus:


Xefyr0:
These lang entries belong to GregTech: Modern and a PR should be made on their repo instead of Monifactory.

PR #1413, opened 2025-01-14
Jetpacks and thrusters

mustapelto:
Updated textures for jetpacks and thrusters.


tomprince:
At a glance, the new thruster textures look fairly easy to confuse with gas buckets (which are upside down.

mustapelto:
> At a glance, the new thruster textures look fairly easy to confuse with gas buckets (which are upside down.

You're absolutely right, don't know how I didn't notice this. Fixed.

PR #1431, opened 2025-01-19
Revert "Change T4 rocket material back to Cryolobus (#1427)"

Xefyr0:
This reverts commit 78308e5823745398f0e2d142a73d60e9bd2123f5.

OK, this rocket has successfully tricked me.
As previously mentioned, Dark Soularium, the material used in the thrusters of the rocket, requires Trinium coils.
I've confirmed this to be true even with with LuV and ZPM voltage boosts to EBF temperature.

However, these Trinium coils in turn require a ZPM assembler. This means that the T4 rocket is mid-ZPM, regardless of whether the plating is Cryolobus or Tritanium. Therefore, all of the balancing I've been trying to do goes out the window, and it's probably better to have it be Tritanium plating to match the default color since it doesn't make a difference anyways.


Xefyr0:
OK, I'm fucking cooked.
The ZPM assembler requires some ZPM components and circuits. This can all be setup _without_ a research station, and without having to set up Crystal Circuits. This means that Dark Soularium can come before those two milestones. Tritanium, being gated behind Fusion II, cannot. 

With all this in mind, Tritanium is much later in ZPM than Dark Soularium (and Cryolobus!) are and therefore it would be beneficial to stick with Cryolobus as the rocket's plating.

Fuck me.

PR #1439, opened 2025-01-24
Stonecut GTCEU warning blocks into each other

gamma-delta:
For parity with Chisel.

I like the textures but can't be arsed to fiddle with black and yellow dyes all day.

### DIFF
--- a/kubejs/server_scripts/random_recipes.js
+++ b/kubejs/server_scripts/random_recipes.js
@@ -604,7 +604,17 @@ ServerEvents.recipes(event => {
             });
         });
     });
-
+    
+    // Stonecut GTCEU warning blocks into each other
+    let hazardNames = ['acid', 'antimatter', 'bio', 'boss', 'causality', 'explosion', 'fire',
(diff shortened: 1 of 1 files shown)


Xefyr0 (changes requested):
23 inputs x 23 outputs for all possible stonecutting recipes is a lot of new recipes for a slight improvement in QoL. 529 recipes! 1/4 of all stonecutting recipes would be hazard blocks after this change!

Perhaps it would be better to do something similar to P2Ps, where one craftable warning block can be stonecut into all of the others.

gamma-delta:
Ah, I can make the "root block" the solid machine casing, which you can cut to and from any warning block, would that be better!

ThePansmith:
I feel like this would be something better directed towards GTm, as it's still getting actively updated.

But yeah for now just do what xef said.

gamma-delta (commented):
Yup

gamma-delta (commented):
Aha sorry I am not very good at the Github review workflow. hopefully this should review it properly?

Xefyr0 (approved):

PR #1443, opened 2025-01-25
Proof-of-concept Canner recipe for Infinity cells.

tomprince:
This depends on KubeJS-Mods/Rhino#51.

It is also fairly messy and could probably be made simpler.


WithersChat:
This can probably be closed since we're killing tank.

Xefyr0:
Tank kill is imminent and we have a coremod now. Sorry Tom, this isn't needed anymore.
The technique is cool tho, I'll try to remember it in case I decide to do some cursed schenanigans in KubeJS :trollface:

PR #1446, opened 2025-01-26
Ultimate Gem Rework

SparkedTheorem:
Currently the Ultimate Gem is just smashing a bunch of gems + some random ones that had zero use at any other point in the game into an extended crafting table to create an endgame component. This is extremely boring, so it should be reworked into an actual processing line instead of "oh yeah I need to smush these things together in an auto extended crafting table".

Some other materials such as Endest Star and Ultimate Core have had gregified recipes in order to make them more interesting as components of the Ultimate Gem.

Also intend to remove some of the more random crystals like Cinnabar.


SparkedTheorem:
Ultimately deciding to go ahead with the idea of just removing Ultimate Gem and Ultimate Core.

PR #1458, opened 2025-01-29
Mana and ABS

### LINKED ISSUE #1388: [Feature]: ABS Mythril recipe
### Suggestion

Mythril is an alloy of Titanium dust and Primal Mana dust, it would be nice to give players the option to make it in **Alloy** Blast Smelter.

iluha168:
Moved Primal Mana Dust and Primal Mana to GT registry.

- Adds ABS recipes
  - microversium, manyullyn, mythril (Closes #1388).
- Potential to auto-generate thermal ABS recipes in the future.

This PR does not affect balance, all ratios are the same.


tomprince (changes requested):
I'm not sure that this is the right approach. It seems you are fighting GT places.

In addition, this will void all primal mana from existing worlds.

Issue #1465, opened 2025-01-31
[Bug]: MISSING Uraninit Ore


FNKM625:
### Issue Summary

NO, uraninit ore generation on 0.11.4
seed:2656459171796697909
with uraninit and pitchblend in end it`s ok

### Reproduction steps

 
1) do backup
2) launch save on 0.11.3 and find uraninit
2.1)![Image](https://github.com/user-attachments/assets/5082ab9a-adc3-4c09-b80f-39f827c80203)
3) swap version and launch on 0.11.4
4) find same place
4.1)![Image](https://github.com/user-attachments/assets/19ae39fe-b834-42a4-a2be-ac7d843f1c6b)
5) no uraninit, who stolen my uraninit in overworld?

### Relevant log output

```shell

```

TheHalfPotato:
Got the same issue, prospector shows 0 uraninite anywhere - Me and a friend have been grid searching several thousand chunks for 3 hours now.

However turning on GT nodes in journeymap shows the uraninite veins all over the place, digging out the entire chunk from top to bottom reveals 0 pitchblend and uraninite.

talchas2:
I don't see any default nonzero value for density in GTOreDefinition or friends, so I bet the #1425 0.5 -> sqrt(density()) change made it so that all of the veins defined by moni in oreVeins.js don't work, because none of them specify a density, and relied on the default 0.5 value forced by that modifyAll.

mgrosven:
I set a default density in `kubejs/server_scripts/worldgen/oreVeins.js` and veins are spawning again. This appears to be the same issue as #1464.

branan:
I'm the filer of #1464. I closed it in favor of this one since they are indeed the same issue and there's more discussion here.

@talchas2 and @mgrosven are correct, as best I can tell. Obviously `sqrt(NULL)` is `0` 🙃. I love javascript.

I can confirm that adding a new default density fixes things as @mgrosven said.

My best guess at an overall fix would be to set all the custom veins to a default density of `0.25` - that gives them the previously intended density of `0.5` when the new sqrt logic is taken into account. It's midnight in my timezone, but if I don't see a PR when I'm awake in the AM, I'll go head and open it myself.

PR #1466, opened 2025-02-01
Fix Custom Ore Vein Worldgen

mgrosven:
Set a default density of `0.25` on custom ore veins so when `sqrt` is applied the previous(desired) value of `0.5` is set.

This will fix issue #1465.


MidoriKami:
image

Applied this fix manually, and can confirm that at minimum tetrahedrite is spawning again, lmk if you want additional testing/pics

PR #1469, opened 2025-02-02
Replace first half of Nethline with standardized ore processing

Xefyr0:
Title.
The first half of nethline seems to be a fascimile of regular ore processing, but with very high energy costs.
With Nethline being as slow as it is, this would increase the speed at which nethline can be performed while making it consistent with other types of ore processing - including the "Ore Processing Diagram" page other ore types have.

PR left as draft until Withers, the original author of the Netherite line, can review.

nethore


WithersChat:
Go ahead, that looks solid to me. I'll *probably TM* rework the last step.

PR #1471, opened 2025-02-03
Use GT materials for dense ore decomposition.

tomprince:
Also adjust output amounts to account for original ore type.

### DIFF
--- a/kubejs/server_scripts/gregtech/dense_ores.js
+++ b/kubejs/server_scripts/gregtech/dense_ores.js
@@ -3,22 +3,22 @@
  */
 
 const ores = [
-    {name: 'redstone', mod: 'gtceu'},
-    {name: 'diamond', mod: 'gtceu'},
-    {name: 'emerald', mod: 'gtceu'},
-    {name: 'lapis', mod: 'gtceu'},
-    {name: 'coal', mod: 'gtceu'},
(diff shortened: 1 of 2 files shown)


kubejs/server_scripts/gregtech/dense_ores.js
```diff
@@ -3,30 +3,30 @@
...
+            .itemOutputs(ChemicalHelper.get("raw", name, amount))
             .inputFluids(Fluid.of("gtceu:nitric_acid", 1000))
             .duration(200)
             .EUt(500)
     })
 
     event.recipes.gtceu.chemical_reactor("kubejs:undensify_oilsands")
         .itemInputs("kubejs:dense_oilsands_ore")
-        .itemOutputs('16x gtceu:raw_oilsands')
+        .itemOutputs('16x gtceu:endstone_oilsands_ore')
```
tomprince, line 29: I had left this untouched when I was doing the change, to allow the use of the centrifuging recipe for oilsand blocks. It allows using the miner trick for it, but oil is cheap enough that I don't think it maters.

tomprince (commented):
I've had this sitting around for a couple of days, and think it makes more sense to use GT materials for the outputs, rather than needing to encode the mod from which the raw ore comes from.

PR #1479, opened 2025-02-03
Microminer Rework - repair edition

Xefyr0:
Placeholder draft PR as a reminder that this is a thing that _also_ exists.

**Current state:** In need of feedback & testing.

Potential future additions:
- Reconsider microminer fuels
- Tweak microminer recipes to not use so much Tungsten Carbide everywhere
- Tweak microminer recipes & textures to match one another
- Make T7MM plating not just cryolobus spam
- Add more missions for ores that can be gotten with Monicoins but not microminers
- More interesting actualization chamber multiblock structure, or remove Stab miners entirely
- Replace Dense ores with equivalent amount of raw ores


Xefyr0:
Ignore this one - replaced with #1480

PR #1482, opened 2025-02-03
tweaks to MM recipes

SparkedTheorem:
reduces overall spam of cryolobus and tungstencarbide in later MMs (decided to keep tungstencarbide in T4MM since its like 4 double plates)

-T7 now uses HSS-S in the center areas and on parts of the wings
-T8 now uses Duranium in place of Tungstencarbide, helps to better intertwine normal GT progression and Nomi unique progression
-T4.5MM now uses HSS-G in place of Tungstencarbide

### DIFF
--- a/kubejs/server_scripts/microverse/miners.js
+++ b/kubejs/server_scripts/microverse/miners.js
@@ -121,23 +121,24 @@ ServerEvents.recipes(event => {
 
     // T7 Microminer
     event.recipes.extendedcrafting.shaped_table('kubejs:microminer_t7', [
-        ' L  D  L ',
-        ' DDHGHDD ',
-        'LDDHHHDDL',
-        'DHDHPHDHD',
-        'DHDAEADHD',
-        'DDDHPHDDD',
-        'D DHHHD D',
-        'D DHDHD D',
+        ' L G G L ',
+        ' D DDD D ',
+        'LDNDDDNDL',
+        'NDDFPFDDN',
+        'DNDAEADND',
+        'DNDNPNDND',
+        'DNNDFDNND',
+        'D NDDDN D',
         'D WW WW D'
     ], {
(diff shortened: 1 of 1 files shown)


Xefyr0:
Out of curiosity, why choose HSS-S for the T7 and HSS-G for the T4.5MM?
I agree the Duranium would be a good addition for the T8, but I think it would look better and not hurt the preexisting Duranium bottleneck as much if it were a replacement for the Iridium, which you will already need a fair amount of for the T5MM. Instead, I think RPP could be used to replace the Tungstensteel:
t8mm

I also am partial to the idea of using Naquadah as the "alternate" material for the T7MM, and slightly tweaking the recipe shape to add Enderium miner cores and to match the sprite better:
t7mm

What do you think?

SparkedTheorem:
> Out of curiosity, why choose HSS-S for the T7 and HSS-G for the T4.5MM? I agree the Duranium would be a good addition for the T8, but I think it would look better and not hurt the preexisting Duranium bottleneck as much if it were a replacement for the Iridium, which you will already need a fair amount of for the T5MM. Instead, I think RPP could be used to replace the Tungstensteel: t8mm
> 
> I also am partial to the idea of using Naquadah as the "alternate" material for the T7MM, and slightly tweaking the recipe shape to add Enderium miner cores and to match the sprite better: t7mm
> 
> What do you think?

RPP could be a good replacement for Duranium if its bottleneck is too bad. I thought about RPP at first but I thought that tying T8MM to fusion would be more interesting.
Regular Naquadah is probably fine for the T7MM now that I think about it, initially I thought regular Naquadah might be a little too plentiful and that Osmiridium might be a bit too expensive so HSS-S was a decent middle ground.
HSS-G on the T4.5 I thought of as a progression middle ground between Tungstensteel (start of IV) and Iridium (mid-late IV).

PR #1484, opened 2025-02-04
Usable complex SMDs 

iluha168:
Made complex SMDs usable as a substitute for 4 advanced SMDs.

### DIFF
--- a/kubejs/server_scripts/gregtech/circuits.js
+++ b/kubejs/server_scripts/gregtech/circuits.js
@@ -72,6 +72,110 @@ ServerEvents.recipes(event => {
             .EUt(245760)
     }
 
+    for (const javaRecipe of event.findRecipes({ mod: "gtceu" })) {
+        /** @type {import("../../dx/typings/GTJSONRecipe.d.mts").GTJSONRecipe} */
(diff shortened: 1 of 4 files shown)


kubejs/server_scripts/gregtech/circuits.js
```diff
@@ -72,6 +72,110 @@ ServerEvents.recipes(event => {
...
+                if (c.type !== "gtceu:sized" || "type" in c.ingredient)
+                    throw new Error("Cannot generate complex SMD recipes")
+                let ing = c.ingredient
+                return {
+                    id: "tag" in ing? "#"+ing.tag : ing.item,
+                    amount: c.count
+                }
+            })
+        )
+        let [newInputFluids, newOutputFluids] = [recipe.inputs.fluid, recipe.outputs?.fluid].map(items =>
```
tomprince, line 114: I *think* there is probably a much less verbose way to implement this, using `GTRecipe.copy`, and modifying the resulting recipe. I think the only bit that isn't straightforward might be adding the resulting recipe to the game.

This avoids needing to know how to replicate all the details of the existing recipes.

PR #1491, opened 2025-02-06
More QB changes

Jollahs:
Sorry for so many individual commits, I had trouble doing stuff at first so I went one at a time.

Also earlygame.snbt claims a quest got deleted related to numis, I found nothing, let me know if I broke something.


Xefyr0 (changes requested):
Two teeny nitpicks:
1. T5MM is not the _only_ source of Naquadah now that Glacio is a thing. It is, however, the first source of Naquadah available to the player.
3. The T5MM Naquadah mission gives Pollucite in place of Sheldonite for some element (I think Cerium?) needed in Snowchestite processing

Xefyr0 (approved):

PR #1496, opened 2025-02-07
tweaks to MM recipes

Xefyr0:
Accomplishes what #1482 was supposed to do, but on the remaining microminer rework branch and with a little extra.


Xefyr0:
Whoops, mixed up my branches again :skull:

PR #1503, opened 2025-02-09
Gitignore rework

iluha168:
- [x] Deduplicates idential config files in the repo
- Adds dev-community-proposed gitignores 
  - [x] Pack switching
  - [x] Probe JS v7


Xefyr0 (changes requested):
Please find a way to do this that doesn't involve copying all of the files to/from config overrides in this commit and every build/mode switch. That's over 300 files to modify every time.

Xefyr0 (approved):

PR #1514, opened 2025-02-16
Fix Tantalum and Pyrochlore fluorine imbalance

Jollahs:
each recipe uses 4B HF, not two. Reaction is balanced


Xefyr0 (changes requested):
The Hydrofluoric acid requirement for Tantalite doesn't need increasing like with Pyrochlore. Instead, the Manganese Difluoride output should be halved to balance out both Fluorine and Manganese.

Xefyr0 (approved):

PR #1512, opened 2025-02-15
Alternate AE2 recipes for printed silicon, again

Edgaru089:
Pulls in the GregTech Wafers recipe for AE2 Printed Silicon from 29364c3 by @Xefyr0, which was missed in fa61955 pulling most of #1200.

Silicon, phosphorus, naquadah, neutronium and universe wafers make 1, 2, 4, 8 and 16 printed silicons each in AE2 inscribers and GregTech forming presses.

PR #1515, opened 2025-02-17
Alternate recipe for AE2 printed silicon in Expert Mode

Edgaru089:
Reopen of #1512, sorry for the mess.

Pulls in the GregTech Wafers recipe for AE2 Printed Silicon from 29364c3 by @Xefyr0, which was missed in fa61955 pulling most of #1200, and makes said recipes only in Expert Mode.

Also disables the AE2 silicon recipe in Expert Mode.

Silicon, phosphorus, naquadah, neutronium and universe wafers make 1, 2, 4, 8 and 16 printed silicons each in AE2 inscribers and GregTech forming presses. The problem with more output is that, universe wafers would make more than 64 silicons per recipe in gregtech forming presses, and it wouldn't work cuz there's only one output slot (can we have more?)

### DIFF
--- a/kubejs/server_scripts/mods/AE2.js
+++ b/kubejs/server_scripts/mods/AE2.js
@@ -403,21 +403,51 @@ ServerEvents.recipes(event => {
     }).id("kubejs:ae2/calculation_processor_print")
 
     event.remove({ id: "ae2:inscriber/silicon_print" })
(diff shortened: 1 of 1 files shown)


Edgaru089:
There you go, fixed linting errors.

PR #1518, opened 2025-02-18
Sorry, I dropped the microminers.

WithersChat:
- Fixed some colors on MMs
- Also removed useless metadata from some MM textures
- Added textures for damaged miners


WithersChat:
I lied. I also removed damaged MMs that won't be used.

DISCLAIMER: THIS PR IS NOT BALANCED, MERELY SETTING THE BASE FOR BALANCE TO BE IMPLEMENTABLE AND ADDING TEXTURES.

PR #1519, opened 2025-02-18
Change actinium electrolysis into its own mini-line like Indium

Xefyr0:
Side additions:
1. Actinium has 4x higher yield through both electrolysis and fission
2. Thorium is now more grey, to better match the NCN fuel pellet
3. Neutron Emitter recipe is 1/4 as long


WithersChat:
Seems good as long as we have reliable sources of the ores used.

PR #1527, opened 2025-02-25
Generate UUIDs deterministically

9ary:
This ensures generated files are stable (byte for byte/same hash) across rebuilds, which improves reproducibility.


9ary:
Oops. Lint should pass now.

Xefyr0:
We went with the implementation in #1608 to use the UUID package instead.
Thank you for the suggestion, though.

PR #1534, opened 2025-02-27
Better Regolith Dust Loot Pools

jude123412:
I felt like something was missing with the Planetary Ores, mainly how little Regolith dust you get from them compared to regular GT Ores, So I fixed them.

Now all Planetary stone types will drop an approximate amount relative to how much dust is dropped from GT ores. 

I have Also tweaked the drop tables to give less Regolith dusts when in hard or expert mode.


Xefyr0 (changes requested):
This does not approximate GregTech ores' dust drops any better than the current state of things, and GregTech ores do not change frequency of drops depending on difficulty.

I suggest you look at lines 220-232 of https://github.com/GregTechCEu/GregTech-Modern/blob/1.20.1/src/main/java/com/gregtechceu/gtceu/core/MixinHelpers.java
for reference.

jude123412:
Ah, I get how it works now, I was basing the code on average drops of GT Ore blocks. I'll make the changes soon :)

jude123412:
This should be the final commit for this PR. Ill be waiting on it to be reviewed :)

Xefyr0 (approved):

PR #1551, opened 2025-03-09
Packmode Overhaul

Xefyr0:
Before this PR, most mode-specific functionality (Content, quests, and otherwise) was selectively enabled or disabled based off of a few mutually exclusive KubeJS variables that depended on the difficulty/mode the pack was set to: Normal, Hard, or Expert.

This adds an additional abstraction layer between the mode and the enabled/disabled functionality, which dramatically eases the process of adding new pack modes or altering existing pack modes.

To do so, the mode-based KJS variables are swapped with feature-based KJS variables - e.g. a boolean that toggles LaserIO nodes and item & fluid cards, one that toggles Monicoin-related content, another that toggles steam age-related content.
These variables also toggle related quests.

Unfortunately, there is no guarantee that any given permutation of variables will be error-free, especially given how modes tend to toggle certain configs as well. (namely `gtceu.yaml`) Official modes have been tested, but a warning is put in-place for any players who choose to modify the scripts to create their own experience.


ThePansmith:
Merged via packmode

PR #1563, opened 2025-03-18
[MMRework] Remove last of dense ores

MidoriKami:


MidoriKami:
Note: latest commit adds a new mission to split Diamond/Emerald and Almandine/Topaz/Sapphire

PR #1567, opened 2025-03-21
Give flux recipes unique recipes ids so that they don't shuffle around and cause emi / ae2 issues

cubebotfan:
For some reason, the flux recipes don't always get given the same recipe ids, sometimes they'll trade ids around between relogs which causes issues with EMI and AE2. This pull request fixes that.

This pull request also adds mixer versions of the quantum flux crafting recipes for people who want to automate it using gregtech machines instead

### DIFF
--- a/kubejs/server_scripts/aa_planet_resources.js
+++ b/kubejs/server_scripts/aa_planet_resources.js
@@ -30,7 +30,7 @@ ServerEvents.recipes(event => {
         ], {
             A: planetResources[2],
             B: `kubejs:${planetResources[1]}`
-        })
+        }).id(`kubejs:quantum_flux_from_${planetResources[2]}`)
     })
 
     // Rock dust centrifuging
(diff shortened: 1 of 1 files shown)


cubebotfan:
I'm going to do a quick check to see if any other recipes have similar issues to this and will bundle those into this pr if I find any

cubebotfan:
Nevermind, I don't have a good way of checking recipe ids for duplicates, we just have to fix them as we find them

kubejs/server_scripts/aa_planet_resources.js
```diff
@@ -30,7 +30,7 @@ ServerEvents.recipes(event => {
         ], {
             A: planetResources[2],
             B: `kubejs:${planetResources[1]}`
-        })
+        }).id(`kubejs:quantum_flux_from_${planetResources[2]}`)
```
MidoriKami: change to `planetResources[1]` instead, as `planetResources[2]` adds in colons into the recipe path making it invalid.

image

cuberyl-catastrophe: That's what I get for not testing, it should be fixed now

Xefyr0:
y'all half of this is already covered in #1547 :skull:

cubebotfan:
Makes sense to close this now then

Issue #877, opened 2024-10-07
[Feature]: Greenhouse (or new multiblock) Item-Based Mob Breeding


abculatter2:
### Suggestion

For mass-producing meat in a more lag-friendly way, particularly for Hardmode players.

Currently, we have soul vials from EnderIO already, so all it'd really require is the recipes.

Suggested recipes:
2 animals (not consumed) + 2 breeding item + 5 minutes @ 8eu/t = 1 animal
4 animals (not consumed) + 4 breeding item + 5 minutes @ 16eu/t = 2 animals
6 animals (not consumed) + 6 breeding item + 5 minutes @ 24eu/t = 3 animals
8 animals (not consumed) + 8 breeding item + 5 minutes @ 32u/t = 4 animals

The recipe can also have speed potion fluid (or potion items?) added to it to halve the time, and healing potion can be added to double the output and/or allow more then 8 animals per process.

Also, the 8eu/t number is arbitrary, though the 5 minutes is based on the breeding cooldown, in case that wasn't obvious.

Once you have the soul-vial animals, they can be fed into various machines to give different drops;
Cutting machines give normal drops, as if killed with a normal unenchanted sword
Arc Furnace gives cooked drops
Macerator gives mince meat and bonemeal
Extractor (or some other machine?) gives experience juice, possibly requiring something like a vibrant crystal with a 1% chance of being consumed as a "catalyst".

These could also feed into other recipes. For instance, putting meat into a fermenter can give rotten flesh (good for more leather), or you could add a more complex but cheaper resource cost version of growth medium and/or bacteria that uses a variety of foods in a mixer.

LCoolJT:
this could also be done by adding some of the simulation data back into the game for cows etc.
(Hostile Neural Networks)

abculatter2:
Hardmode removes HNN entirely, and I feel that mob breeding is not a particular powerful thing that would affect the balance of hardmode, so it would be preferable to make it not rely of HNN.

PR #1580, opened 2025-03-25
Add a way to reprocess Rotten Flesh into Mincemeat and Fermented Biomass

Xefyr0:
This adds a short 2-step process that turns Rotten Flesh, Distilled Water, and Iodine dust into Tiny piles of Bonemeal, Fermented Biomass, and Mince meat.

It would serve as an alternative to grinding up copious amounts of Fish to get enough Mincemeat for Wetware.
It would also cover the need described in #877.


WithersChat:
Sounds great honestly. What are sources of rotten flesh in HM tho?

PR #1585, opened 2025-03-27
Mega Alloy Blast Smelter

Xefyr0:
Finally made a design people might accept.
No longer depends on blocks from #1344.


WithersChat:
Seems good to me.

PR #1597, opened 2025-04-02
Add CubeJS joke item

cewlboi:
blame withers :trollface: 

Recipe uses Omnic data even though tank kil has not been implemented yet at the time of writing
kjs_cubejs


ThePansmith:
:trollface:

PR #1593, opened 2025-03-30
Transition to the nukeList paradigm for removing items & fluids

Xefyr0:
Before this PR, a large quantity of items & fluids are separately hidden from EMI/JEI, and recipes/tags removed from them.

This PR replaces those distinct operations with a single one, where all items and fluids that should be effectively removed from the game go on a "NukeList" where they are hidden from EMI/JEI, tags removed from them, and all recipes that use them are removed.

This ensures uniformity across "removed" items and fluids and simplifies the process.

PR #1603, opened 2025-04-04
Deprecation Pipeline

Xefyr0:
As development continues some items, and fluids may need to be deprecated in favor of newer versions.
This PR implements utility methods to assist in that process, and transitions a few items & a fluid currently in-process of being deprecated to this new paradigm both for neatness and as an example:
- Exoskeleton Legs
- Molten Primal Mana
- Dilithium Crystal

Functionality for blocks NYI - hopefully it follows similar patterns when such functionality may be needed in the future.


WithersChat:
Thinking of merging this and #1593 after 0.12 once we'll have the time to double check if anything breaks.

PR #1608, opened 2025-04-05
Deterministic build

iluha168:


tools/build/codegen/credits/target.js
```diff
@@ -134,8 +136,8 @@ export const CodegenCreditsTarget = new Juke.Target({
                         "{{Y_BUTTON}}": y + hoverOffsetY,
                         "{{IF_HAS_LINK}}([^]*?){{IF_HAS_LINK_END}}\n": data.link ? (_, match) => match : "",
                         "{{LINK}}": data.link ?? "",
-                        "{{BUTTON_EXEC_ACTION_UUID}}": randomUUID(),
-                        "{{BUTTON_EXEC_BLOCK_UUID}}": randomUUID(),
+                        "{{BUTTON_EXEC_ACTION_UUID}}": uuid5("ButtonExecAction", uuidNamespace),
```
9ary, line 139: Doesn't this result in many duplicate UUIDs?

iluha168: Nope

iluha168: `uuidNamespace` is unique as long as the input data for a contributor is unique

9ary: Oh, I missed that bit. My bad.

PR #1609, opened 2025-04-05
Remove modifier bonuses from Exoskeletons

MidoriKami:
The way that the item targeting works here causes this script to stack its effects with the new gtexoskeleton mod resulting in doubled speed bonuses and doubled height bonuses.


ThePansmith:
also just to get people to swap over

PR #1615, opened 2025-04-06
More sensible LCR Sulfuric Acid

MidoriKami:
Reduce LCR Sulfuric Acid to 2s per operation at HV

Reasoning: Sulfur Trioxide 3-part recipe at HV takes 1.6s per 1b Sulfuric Acid due to perfect OC, but since the LCR recipe is based at HV it takes 16s to do the whole operation.

*Alternatively*, if we want to encourage players to consider making the 3-LCR process, we can modify this change to bring it down to 4s instead of 2s.

### DIFF
--- a/kubejs/server_scripts/_hardmode/hardmode_processing.js
+++ b/kubejs/server_scripts/_hardmode/hardmode_processing.js
@@ -274,7 +274,7 @@ ServerEvents.recipes(event => {
             .inputFluids("minecraft:water 4000")
             .circuit(24)
             .outputFluids("gtceu:sulfuric_acid 1000")
-            .duration(160).EUt(480)
+            .duration(40).EUt(480)
 
         // Fluoroantimonic Acid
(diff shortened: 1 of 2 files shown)


Xefyr0:
You're forgetting an important cost of the Sulfur Trioxide recipe: the Oxygen requirement. 3B are needed per bucket of Sulfuric Acid, which changes the quantity of machines needed for the Trioxide route from 3 to 12 if you use Silicon Dioxide for Oxygen production and still want to produce at one bucket per 1.6s. The C24 recipe skips the need to make Oxygen, and uses Water directly. It's a _huge_ simplification.

Granted, most players have a robust centralized Oxygen production somewhere in their base that they can leech off of, but it's still 3 machines to 1 in that case and frees the Oxygen for use elsewhere.

MidoriKami:
Fair enough, thanks for the thorough explanation.

PR #1617, opened 2025-04-06
Make Fluxbore HV

MidoriKami:
Removes fluxbore from expert mode, yall got drills at home already.

Moves fluxbore from *early* MV, to HV

Reasoning being, it's trivial to upgrade fluxbore to 9x9x1 the moment you make one, so it should be balanced between the MV drill (3x3x3) and HV drill (4x4x4)


WithersChat:
I like this.

WithersChat:
Can you update to use new packmode? (Either use whatever boolean makes sense or make your own.)

MidoriKami:
Will do in a couple hours when I get home

MidoriKami:
Was having git issues, move to #1639

PR #1630, opened 2025-04-09
Completely remove ExtendedAE Silicon Block

MidoriKami:


Xefyr0:
It would probably be easier and cleaner to do this with/in #1593.

MidoriKami:
Sounds good to me, get it merged before 0.12.2 and we can close this one =D

WithersChat:
Gonna simply add it to the nukelist instead. Thanks still ^^

PR #1642, opened 2025-04-10
Add AE Nukes

MidoriKami:
;) gotcha with the title.

This PR moves removed AE2 and ExtendedAE items to the nukelist.

I'm not super confident on the cable removals, might be worth checking.


WithersChat:
I think this is a bad plan. Nukelist removes all tags and hides from JEI. But we don't want to remove most of those items, just their recipes (so we can replace them with our own).

Also, the nukelist takes item IDs, not recipe IDs (which you supplied).
In other words, you just deleted AE2 storage cells from the pack. Except that you didn't because recipe IDs don't work in the nukelist.

PR #1648, opened 2025-04-10
Germanium and Advanced Solder Additions

SparkedTheorem:
Adds Germanium and Advanced Solder.

Germanium is obtained from Coal Fly Ash, which replaces Ashes/Dark Ashes in a all recipes where Coal, Coke, and Charcoal made them.

Advanced Solder, made from Bismuth, Tin, Zinc, and Germanium, is available starting at HV, being a more efficient version of regular Soldering Alloy, and replaces all use for Soldering Alloy by LuV. Unfortunately, my brain is too small to make new versions/replace regular Solder with Advanced Solder for all these recipes, so I am letting someone else do it.


SparkedTheorem:
Ok, so as it turns out the current method I use to replace all instances of soldering alloy in the assembly line with advanced soldering alloy straight up does not work until you do /reload at least once. I am not sure how to fix this, and I do not want to try my hand at attempting to weave in the software wizardry that is superconducting_coils.js's method.

PR #1656, opened 2025-04-12
Rework Terbium Line

Jollahs:
fixes stoich
adds oxalic acid and related chemicals to make it


Xefyr0 (changes requested):
1. Ammonium Oxalate from Oxalic Acid and Ammonium Carbonate requires Circuit 5, but this does not display in the singleblock chem reactor view in EMI - only the LCR.
2. I think the Rare Earth Oxide step in the EBF should output Ammonia as a byproduct to recycle. Your call on how much.
3. Please move the Rare Earth Salt + Ammonium Nitrate -> Terbium Salt recipe to another machine to avoid conflicts with the Ammonium Nitrate electrolysis recipe. A Centrifuge or CR should work and you can add a solvent of your choice for the dusts to react in, if you feel it necessary.
4. Oxalic Acid Dyhydrate has too many mols per item - maybe multiply its quantity in every recipe it's in somewhere between 2x and 4x.
5. The Sucrose Solution & Sucrose feel redundant when Sugar and NCN liquid Sugar exist. NCN Liquid sugar will probably be replaced with a GT fluid when NCN gets yote, but either way you're adding what feels like duplicate fluids. If GT labelling Sugar as Glucose annoys you so much then change it's formula and remove or change its electrolysis recipe.

Xefyr0:
As a side comment, *please please please* never again add content that depends on or "fixes" someone else's PR. I know I'm somewhat hypocritical for saying this given how MABS relied upon some blocks from the ICE PR, but in the end I changed it back and unlike that case, the two PRs aren't made by the same person.

iluha168:
@Xefyr0 don't worry, I'll just cherrypick as soon as I can

PR #1658, opened 2025-04-12
Buff parallel sculk core charging

iluha168:


Xefyr0:
Could you change the QB entry to note that the Biobalance charger isn't just faster, but is also 4x more energy-efficient than the regular Sculk Charger?

Jollahs:
> Could you change the QB entry to note that the Biobalance charger isn't just faster, but is also 4x more energy-efficient than the regular Sculk Charger?

added this in #1656

Xefyr0 (approved):

PR #1664, opened 2025-04-12
Add Architect's Palette Compatibility

lilpaladin1:
This PR adds a compatibility script for Architect's Palette so that it works with GregTech machines like having the various Warping recipes be in the Atomic Reconstructor instead because Warping is not possible in this pack.

There's also other changes as well, mainly adding alloy smelter, mixer, compressor, and assembler recipes.

I did want to add a monicoin recipe for Chorus Fruit for Oracle Jelly, but that does ungate warpstones in Normal Mode, so that's not really my call to make.


Xefyr0:
Please make the PR add Architect's Palette to the list of Addon Mods (AKA Optional Compats) in the README:
[](https://github.com/ThePansmith/Monifactory?tab=readme-ov-file#addon-mods-aka-optional-compats)

Xefyr0 (changes requested):
Please make the PR add Architect's Palette to the list of Addon Mods (AKA Optional Compats) [in the README](https://github.com/ThePansmith/Monifactory?tab=readme-ov-file#addon-mods-aka-optional-compats)

Xefyr0 (changes requested):
Got a few more tidbits of feedback now that the formalities are out of the way:
1. When deconstructing blocks into item components (Like Unobtanium, Withered Bone Block -> Withered Bone) use the Forge Hammer instead of the Extractor.
2. Wardstone Blend doesn't have a recipe. It's fine if it shouldn't, just remember to hide it in EMI/JEI.
3. Plating block is *too* cheap. Architect's Palette is generous but 1 nugget for 8 blocks is insane.
4. Withered Bone has no recipe. AFAIK this is because it's normally dropped by Wither Skeletons, but that may either be bugged or just doesn't show in EMI/JEI. Either way, consider replacing its use in recipes with the KubeJS Wither Bone.

Xefyr0 (approved):

PR #1679, opened 2025-04-15
Reorganize End Game Chapter

PlasmaticVoide:
Shuffled around a bunch of quests in the End Game qb chapter, mostly to fix that one row of 8 quests with overlapping arrows going in different directions but also tried to make everything look good


Xefyr0:
Merged with changes in 7dfc350

PR #1689, opened 2025-04-16
Bauxite Line

SparkedTheorem:
Adds an extended processing line for Crushed/Purified bauxite ores that gives better overall yields than simply processing down into dust.


Xefyr0:
Out of curiosity, why would anyone do this line?
It gives increased yields, but Aluminium, Rutile, and Iron (from Hematite) are free and Vanadium/Gallium can be gotten with a little extra effort in Oreproc or Microminers in HV. (where this line would be placed) 
Unlike this line, however, Oreproc and Microminers are mandatory and so I don't see why players would even try it.

Xefyr0:
Answer, for future reference:
Screenshot 2025-04-19 223801

WithersChat:
Not sure I see the point. Like legit it's just not good enough to justify the effort. +1 to Xef here.

ThePansmith:
Yeah, gonna have to also agree with above.

Xefyr0:
Closing due to the line being not particularly useful.
This will also prevent content overlap, since GTM has plans to introduce a similar line that is both simpler and more rewarding, especially in terms of materials that neither this nor regular Bauxite processing can provide.

PR #1700, opened 2025-04-18
Buff saltpeter MM mission

Jollahs:
end game uses a ton of aerotheum, and you need saltpetah for it


MidoriKami:
Additional context, the amount of direct salt output was a lil bit too much as-well, so I suggested setting all three to 48 Raw Ore each.

PR #1706, opened 2025-04-19
Reorganize Early Game Chapter

PlasmaticVoide:
Quests deleted: Inductor, cap bank, vibrant cap bank
Quests added: sulfuric acid link between sulfur and ethylene
Quests changed: Added inductors to capacitor quest for organization sake
Everything else is dependency shenanigans and moving quests around

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "githubPullRequests.ignoredPullRequestBranches": [
+        "main"
+    ]
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-    "githubPullRequests.ignoredPullRequestBranches": [
-        "main"
-    ]
-}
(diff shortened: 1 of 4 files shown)


Xefyr0 (changes requested):
1. Add Chemical Reactor as a dependency of Sodium Potassium.
2. Change the Inductor's task ID and lang BOTH to use the ID 75E2B7E59A591067 to make use of the existing lang entry & task data FTBQuests saves in players' worlds.
3. Change the Capacitors quest title to read "Capacitors and Inductors" or similar.
4. Add the Sulfur Ore quest in Genesis as a dependency to the Sulfur quest.
5. The top-left of the chapter is a bit too on-grid, making that section look uninteresting. Would you please make it more organic like the rest of the chapter?
6. Polyethylene should be as large as Ethylene.
7. The Pulsating Crystal in the Staff of Traveling now needs Pulsating Alloy instead of Vibrant Alloy, so the quest should be moved to The Factory with the proper change in dependencies instead.
8. You should not remove the quest(s) that point to Capacitor banks as an option without offering an alternative in its place. More importantly, the quest for Basic Capacitor Banks still exists in The Factory. Please add back in the Advanced Capacitor Bank quest, but include the Redstone Flux Cell from Thermal and rename the quest to "RF Storage". Make sure to point out the Holding enchantment and RF Coil augments as ways to increase the block's capacity - I want to make more advanced augments to solidify the Flux Cell as a better alternative in the future.

Xefyr0 (approved):

PR #1712, opened 2025-04-20
Update gtmutils to 1.1.4

NegaNote:
GregTech Modern Utilities version 1.1.4

    Fix PTERB coolant not being re-enabled if only one of the coolant vars was zero
    Add more Jade integration, putting the PTERB's coolant drain in the hover info
    Add tool tags to Omni-breaker, update tooltip
    Fix potential structure issue

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1061,7 +1061,7 @@
       "required": true
     },
     {
-      "fileID": 6440094,
+      "fileID": 6442764,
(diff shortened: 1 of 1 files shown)


Xefyr0 (approved):

Xefyr0 (changes requested):
Needs some fixes in version 1.1.5

PR #1733, opened 2025-04-25
Add P2P attunement tags to extended-pattern items

J4Wx:
Small QOL fix to prevent needing to carry a blank pattern or regular pattern provider for P2P attunement.

Adds the `mae2:p2p_attunements/pattern_p2p_tunnel` tag to the Pattern Modifier, and Extended Pattern Provided (Block and Part).

### DIFF
--- a/kubejs/server_scripts/fixes_tweaks/unification/tags.js
+++ b/kubejs/server_scripts/fixes_tweaks/unification/tags.js
@@ -88,6 +88,11 @@ ServerEvents.tags("item", event => {
     event.remove("forge:storage_blocks/beryllium", "nuclearcraft:beryllium_block");
     event.remove("forge:storage_blocks/graphite", "nuclearcraft:graphite_block");
 
+    // AE2 P2P Attunements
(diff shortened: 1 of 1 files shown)


MidoriKami:
Might it also be reasonable to remove attunement tags from Drills? I assume that's the main cause of accidentally changing the type of a P2P.

ThePansmith:
Yeah, that sounds good, do that then ill merge.

J4Wx:
Unfortunately, I don't think electric-drills, wrenches, screwdrivers, etc, work using a tag - this would need to be a change at the MAE2 level, I believe.

EUP2P attunement uses the GT Capabilities system, specifically `CAPABILITY_ELECTRIC_ITEM`

https://github.com/Monifactory/MAE2/blob/43b29bf55a76cbab6acd3089aa3f33c6281db24b/src/main/java/stone/mae2/integration/GregTechIntegration.java#L25

MidoriKami:
Thanks for looking into it, really appreciate that.

PR #1739, opened 2025-04-29
Update GTMUtils to 1.1.10

NegaNote:
Fix a crash when using dual hatches with the PTERB

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1061,7 +1061,7 @@
       "required": true
     },
     {
-      "fileID": 6474709,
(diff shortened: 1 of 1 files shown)


NegaNote:
Marking it as draft because it turns out this version doesn't work with ME hatches, so it'll probably ALSO not work with other custom hatch types too. Working on it.

NegaNote:
Marking it as ready again because further testing shows I was stupid and using the ME hatches wrong. This version works just fine.

PR #1746, opened 2025-04-30
Update GTMUtils to 1.1.11

NegaNote:
Fixes Jade problems with saying the structure is formed even when it's not (to the extent possible), and removes unnecessary tooltip about avoiding building the PTERB with the terminal.


Spicierspace153:
lgtm

PR #1778, opened 2025-05-07
Replace NC:N with KubeJS Implementation

ThePansmith:
I don't need to explain why, so I'm not going to bother :trollface:

- [x] Fission Reactor Logic
- [x] Fission Reactor Blocks
- [x] Port relevant items and blocks
- [x] Port relevant recipes
- [ ] Balance fission recipes
- [ ] Fix Quests
- [ ] Textures (most items do currently have dev textures, but will need a artpass)
- [ ] Deprecation for NCN stuff


ThePansmith:
Closing for now. Hopefully NCN shouldn't be an issue from here on. 

In the event that this changes, a javascript replacement would work better in any case.

PR #1812, opened 2025-05-29
Streamlining nuclear fission

globalista-dev:
### Proposed changes

- Removed non-GT element isotopes.
- Fission fuels are crafted in a forming press with a cylinder mold
- Only four fuels: 
  - TBU (Thorium-Bred Uranium)
  - LEU-235 (Low-Enriched Uranium)
  - MOX (Mixed Oxide)
  - MAC (Mixed Actinide)
- Spent fuel processing requires two steps:
  - Depleted fuel + HNO3 -(centrifuge)-> Major actinides (U, Pu) + Depleted fuel waste
  - Depleted fuel waste + H3PO4 -(centrifuge)-> Minor actinides + byproducts (only lead for now)
  - Einsteinium added as an item and MAC fission output
  - All fission elements (Ac, Np, Pu, Am, Cm, Bk, Cf and Es) are obtainable through these four fuels

### Reasoning
NCN was cluttered. The isotopes it added were useless outside of fission and added a lot of jankiness, with, for example, U-233 being extracted into U-235. Other in-game reactors of similar type, such as the Naquadah Reactor and Large Naquadah Reactor only have two and one fuel types respectively, so four different fuels for the fission reactor is, in my opinion, not a small amount.

NCN's complexity also created a boring gameplay loop, in which you'd basically construct a pre-defined cube, plop a fuel in to centrifuge the depleted form into new elements. Fuel A -> Element B -> Fuel B -> Element C, etc. Base GT lacks this sort of mindless progression and fission could use with a reduction of this repetitiveness.

The recipe changes aim to both make the outputs fit into the recipes and to introduce a bit of complexity to it. They're very broadly inspired by the first step of PUREX and TRUEX, both of which were not entirely implemented because, simply, there's no need for it.

### Possible changes/todo

- [x] Reprocessing recipes stoichiometry
- [x] Extra byproducts


globalista-dev:
On the topic of extra byproducts, fission doesn't only produce actinides. It also has other elements as possible (and extractable) byproducts. Of course there are the famed radioactive ones, like Cs137, Sr90, etc, but they're out of the question, since they have no use in Moni.

Wikipedia provides a [table](https://en.wikipedia.org/wiki/Spent_nuclear_fuel#Table_of_chemical_data) of possible byproducts of Uranium fission, and of those, the majority are present on the pack. Excluding the liquids and Europium (to make it require fusion), this is a list of possible byproducts, which could be chancedOutputs of specific fuels:
- Y, La, Ce, Nd, Sm -> Rare Earth Dust
- Ru, Rh, Pd -> PGS 
- Nb
- Mo
- Ag
- Cd
- In
- Sb
- I
- Cs
- Ba

If these were to be added, I wouldn't add all of them, maybe one per fuel. The first two seem okay, even though they give the player elements that aren't formed in fission, but again, this isn't made to be realistic.

globalista-dev:
Currently, I'm in favor of keeping Neptunium, Curium, Berkelium, Californium and Einsteinium as just KubeJS items, but they could be moved to regular GT Materials if that proves to be useful to material diversity in the End Game.

ThePansmith:
Like mentioned before, going to be saving this for when we review the nuclear multiblock as a whole after materials work is done.

ThePansmith:
I'm going to close this for now given NCN improvements.

PR #1822, opened 2025-06-02
Fix hard mode formic acid reaction balancing

altalk23:
Hard mode formic acid recipes were unbalanced, so I quickly fixed them.

Changes `CH3OH + CO2 -> HCOOCH3` (one oxygen voided) into `CH3OH + CO -> HCOOCH3`
Changes `HCOOCH3 + 3H2O -> HCOOH + CH3OH` (extra water used) into `HCOOCH3 + H2O -> HCOOH + CH3OH`

### DIFF
--- a/kubejs/server_scripts/_hardmode/hardmode_processing.js
+++ b/kubejs/server_scripts/_hardmode/hardmode_processing.js
@@ -394,12 +394,12 @@ ServerEvents.recipes(event => {
 
         event.recipes.gtceu.chemical_reactor("methyl_formate")
             .itemInputs("gtceu:tiny_sodium_methoxide_dust")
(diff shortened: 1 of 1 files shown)


altalk23:
though if the intention with `3H2O` is to have excess water it might make more sense to have `2H2O` as output instead

esmenard:
The first one should rather be HCOOH+CH3OH -> HCOOCH3+H2O

altalk23:
> The first one should rather be HCOOH+CH3OH -> HCOOCH3+H2O

The current process uses liquid phase methanol carbonylation so i didnt decide to change anything about it just fixed the balance 

Also that one is just the reverse of the second equation so like yeah

esmenard:
From the information I can get, methanol carbonylation makes acetic acid (CH3COOH), not methyl formate (HCOOCH3)

PR #1826, opened 2025-06-04
Re-tier sterile hatch to be late UEV

Xefyr0:
Re-tiers sterile maintenance hatch to be after the bulk of the Bioalloy production necessary for UEV, but before the grind for Creative Chest in UIV.


WithersChat:
Made irrelevant by killing UIV and sterile cleanroom

PR #1829, opened 2025-06-05
Rename Holmium to Necrosiderite

NegaNote:
(also add lang to all mats except OTECO)


ThePansmith:
Now do it with correct capitalization :trollface:

PR #1861, opened 2025-06-17
Recipe tweaks

Xefyr0:
Just some assorted change I thought should be made.
As always, some of these may be more controversial than others so they're separated into distinct commits for easy cherry-picking.

Explanation for the changes I assume will be more controversial:
- Believe it or not, IRL fusion is net-negative on energy on elements heavier than Iron. It's why stars are considered old when they start to fuse it - they're running out of lighter fuel. With that in mind, Americium Plasma being energy-positive should be out of the question. In fact, the only reason it exists is as a GTNH reference and ostensibly to provide a means of synthesizing americium, but the regular molten recipe exists for the latter.
- Boron is quite expensive in Monifactory, actually. It's mostly acquired as a byproduct of ore processing salts. Reducing its proportion in ferroboron to a near-IRL proportion improves both realism and gameplay.
- Advanced drill kits are more expensive than basic drill kits and Thorium is less potent for power than Uranium now that NCN's been updated, so the advanced mission should yield more of it to make the investment worthwhile.
- Though both ores yield Arsenic, In base GTM Cobaltite is far more common than Realgar. Thus, Realgar should be gated later in microminer progression due to being rarer and more valuable. Cobaltite is available as early as the T2 miner, however.


WithersChat:
LGTM

PR #1875, opened 2025-06-23
Kill boost part 1

WithersChat:
- Just missing ancient debris crushing and MM repair rates


ThePansmith (changes requested):
Assuming that the crystal chip recipes are removed  due to not being sure what the values GTm is not gonna use. Just comment it out instead, will make readding them easier

PR #1886, opened 2025-06-28
Modified HNN script to add a PPM recipe with an Autoclave

Apoptosis847639:
Added a way to bulk PPM using PBI and 8 pulsating dust in an autoclave.


ThePansmith (changes requested):
That, and it being kinda random. I understand the desire for higher tier PPM recipe, but this feels a bit arbitrary.

Xefyr0 (changes requested):
Better, but you overcompensated too much.
The duration should be 20% of its current amount but address that and the linting issue and it would look good to me.

Xefyr0 (approved):

PR #1888, opened 2025-06-29
New Energy Core Recipes

PlasmaticVoide:
Liquid Cryococcus to fill cores is a temp placeholder until the svat recipes are made
It was originally planned for the cores to become RF storage items, however I have no idea how to do that, so if anyone wants / knows how to then feel free


Xefyr0 (changes requested):
Energy cores, lore-wise, are what contain the energy that is discharged in the Discharger, so they shouldn't be crafted in there.

With Mesol and Bathyal cores merged, you should probably merge their recipes too.

Lastly, please make sure to double-check for typos by launching your game before making the PR. - `enderio:advanced_soldering_alloy` does not exist.

Xefyr0 (approved):

PR #1898, opened 2025-07-01
Laserio Cards Tweaks

Xefyr0:
Essentially the same as in #1894.
The primary change here is that the wires used in overclockers double in thickness.
Even after this change, it's important to note that the ingots for those wires are still less than 1/3 the total ingot cost for any tier of overclocker. The remaining 2/3 is Iron.
Again, it is also important to point out that LaserIO overclockers can cover arbitrarily large distances through node chaining for the same cost of a single conduit block.

This PR also swaps out the Iron plate in the Fluid Card recipe for Lazurite, because Lazurite is obtainable from the same Overworld vein necessary to make GT fluid filters and matches the blue on the item texture.

### DIFF
--- a/kubejs/server_scripts/mods/laserio.js
+++ b/kubejs/server_scripts/mods/laserio.js
@@ -112,89 +112,31 @@ ServerEvents.recipes(event => {
                 "BBB",
                 "AAA"
             ], {
-                A: "#moni:overclocker_metal_plate",
+                A: "gtceu:iron_plate",
                 B: `gtceu:${value[1]}_double_wire`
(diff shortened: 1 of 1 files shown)


ThePansmith:
Unless if conduits get fixed, I don't see a reason to nerf LIO (and even then)

ThePansmith:
I had a vision. :trollface:

But in all seriousness, remember that each node is far more expensive than a conduit, so it balances out. (if conduits were still in the pack anywho)

Also remember that this would affect balancing for anything that uses said cards in recipes, such as thrusters, jetpacks, etc.

PR #1920, opened 2025-07-06
New Bioalloy Line

PlasmaticVoide:
Completely reworked processing line for Sculk Bioalloy that now uses the new Sculk Vat multiblock


ThePansmith (changes requested):
Doublecheck that all relevant recipes have been cleaned up. Bioalloy block recipe still exists, for example.

PR #1926, opened 2025-07-09
Abbreviate HM platline

Xefyr0:
Vaguely recall reading a message from Jollahs saying that they didn't feel like making PRs to Moni anymore at some point after the fluid laser engraver incident so I did this.

Abbreviated by 6 recipes (+/- 1) and 7 intermediate products (+/- 1)
Mostly me going in with scissors and Krazy Glue and rearranging lines to have fewer recipes but keeping the interesting parts.

Also moves all the platline recipes to a dedicated file.


Xefyr0:
OK it seems that I can't find the aforementioned message, so I may have misread some offhand comment or similar.
Regardless, it's been obvious that HM platline is too much for quite some time now, so this tones it down slightly in the event that Jollahs Platline never comes to be.

PR #1943, opened 2025-07-18
QB Fixes + Misc

PlasmaticVoide:
Some assorted QB fixes
Removed exotic material catalyst and eternal catalyst
Removed duplicate ORB recipe
Removed enderium crystal chip growing recipes
Fixed uev energy hatch  to use hyperdegenerate darconite

### DIFF
--- a/kubejs/server_scripts/gregtech/Post_UV_Components.js
+++ b/kubejs/server_scripts/gregtech/Post_UV_Components.js
@@ -4,8 +4,8 @@
 
 ServerEvents.recipes(event => {
     const converter = [
-        ["uev", "cable", "darconite", "hyperdegenerate_darconite", "1966080"],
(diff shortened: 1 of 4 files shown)


Xefyr0 (approved):

kubejs/server_scripts/gregtech/Post_UV_Components.js
Xefyr0, line 1: Can you undo the changes in Post_UV_Components?
HDD is a superconductor and shouldn't be in energy hatches, and turning numerical values into strings does not play nice when they're used in functions that expect numbers later on.

Xefyr0 (changes requested):
Whoops, turns out I didn't check everything.

Xefyr0 (approved):

PR #1962, opened 2025-07-30
Prismac Recipes Refactor

PlasmaticVoide:
Refactor prismac recipe generation to use loops rather than individual recipes


WithersChat:
Check your null output amounts

Xefyr0:
Merging this, under the assumption that the :fire: react from Withers in #moni-dev is indicative of approval since I have not found any issues.

PR #2000, opened 2025-08-11
Fix build-server hang

DragonDuck24:
Fixes a hang until timeout (5 min) at the end of completing a server build by adding a process exit with code 0.


DragonDuck24:
Don't merge yet, I will try to look into a better fix

DragonDuck24:
Most recent commit should fix the root cause of the issue.

PR #2006, opened 2025-08-13
Fix methyl formate and acetic acid conflict

Jollahs:
cant do palladium in platline without methyl formate :shruge:


Jollahs:
oh right this also adds the new octane recipe cause i forgor

PR #2017, opened 2025-08-15
Give the Crystal mat set's frames CTM.

Flare32:
### I think the title explains it all, doesn't it?
*LDLib? what LDLib? we don't need LDLib!*

Also now fixes a stupid mistake i made in the first crystal set pr.


Xefyr0 (approved):

WithersChat:
I appreciate the attempt but will close this due to technical limitations.

PR #2018, opened 2025-08-16
HNN Training gets Withersnuked :3

Apoptosis847639:
Buh bye HNN training!


Xefyr0:
If you're still having issues, have you considered deleting the non-self-aware model tiers?
and setting the data requirement for the self-aware tier to zero?
so model items never have to gain data (and never can)?

Apoptosis847639:
> If you're still having issues, have you considered deleting the non-self-aware model tiers? and setting the data requirement for the self-aware tier to zero? so model items never have to gain data (and never can)?

pretty sure that breaks things, if you want I can try that again but for now i'll leave it up for review

WithersChat (approved):
Seems legit. Assuming you tested it that is.

Xefyr0 (changes requested):
There's a tooltip on models claiming that they can still be trained in the Simulation Chamber.
More importantly, when I tested out actually crafting a model it gave me the Basic tier instead of the Self-Aware tier.

Apoptosis847639:
> There's a tooltip on models claiming that they can still be trained in the Simulation Chamber. More importantly, when I tested out actually crafting a model it gave me the Basic tier instead of the Self-Aware tier.

Oh crap... I'll try to fix the 2nd issue when I get home, it's a draft now I'll ping you if I want a re-review(I am in Chemistry class rn)

PR #2019, opened 2025-08-17
Tweak half-tier MM repair recipes

cewlboi:
"repairing is cancer" - withers


WithersChat (approved):
Looks good to me. Just waiting on Xef now.

NegaNote:
> Looks good to me. Just waiting on Xef now.

I was misinformed on the XOR stuff, it'll actually work just fine with it included and not introduce any dupe bugs or the like.

Xefyr0 (changes requested):
For T6, 7, 8 miners remove the Thruster, Guidance System, and Cores from the recipes.
Treat the Laser array like the crossbow/sword weapons.
and divide any remaining items' consumption chance by 3 before removing the boosting parameter.

That should make T6-8 miners act like T2.5, T4.5.

Xefyr0 (approved):

PR #2030, opened 2025-08-21
Fix the broken rotten flesh recipe

### LINKED ISSUE #2028: [Bug]: The recipe of rotten flesh is broken
### Issue Summary

The actual crafting recipe of rotten flesh using monicoins is different from the recipe registered in emi.
In kubejs/server-scripts/monicoins/mobs.js line 48, this is the recipe of rotten flesh:
```
            " PP",
            "   ",
            "   "
```
but, in emi. the recipe looks like this:

Image

So, using the recipe registered in emi to craft rotten flesh will result in failure.

### Reproduction steps

 - 

### Relevant log output

```shell

```

bjl32:
Fix #2028

### DIFF
--- a/kubejs/server_scripts/monicoins/mobs.js
+++ b/kubejs/server_scripts/monicoins/mobs.js
@@ -46,8 +46,8 @@ ServerEvents.recipes(event => {
 
         // To jumpstart HNN
         event.shaped(Item.of("minecraft:rotten_flesh", 32), [
-            " PP",
             "   ",
+            " PP",
             "   "
         ], {
             P: "kubejs:moni_penny"


WithersChat:
Did you check for conflicts?

Apoptosis847639:
this is a really weird EMI bug or something.... I think we should look into why EMI showed that if we don't know yet.

bjl32:
> Did you check for conflicts?

yes. there is no conflict.

Xefyr0:
This issue is actually a symptom of emilyploszaj/emi#798. There are several other manifestations of this bug (typically involving Monicoins) Please bug them about it instead if you want to address the root cause.

In the meantime, merging this PR as a band-aid fix.

bjl32:
> This issue is actually a symptom of [emilyploszaj/emi#798](https://github.com/emilyploszaj/emi/issues/798). There are several other manifestations of this bug (typically involving Monicoins) Please bug them about it instead if you want to address the root cause.
> 
> In the meantime, merging this PR as a band-aid fix.

I have checked and tested every single monicoin-related recipes. There is no other recipe issues. Rotten flesh is the only broken monicoin-related recipe.

PR #2037, opened 2025-08-23
Nuke infinity power unit

Xefyr0:
'nuff said.


WithersChat:
Replace it eith largd naq elements in T12MM please but aside from that YES THANK YOU I KEPT FORGETTING!

PR #2038, opened 2025-08-23
Automatically set packmode in builds

DragonDuck24:
This PR primarily sets the pack mode to NM of builds if it has not previously been set.
Along the way, I have also made several other changes to the build script to make this easier (and just improve the script). The notable changes are as follows:

1. Automatically sets built pack to NM (does not affect base repo) if not set previously.
2. `config/`, `config-overrides/`, `defaultconfigs/`, and `kubejs/` have been moved to the root of the **server** build, instead of in `overrides/`. The readme section for the dedicated server installation has been modified to account for this change.
3. Removes the use of a `.tmp/` directory while building, in favor of directly zipping the `dist/{build-type}`
4. `dist/{build-type}` now matches what in in the zip file, meaning:
    1. `dist/client/` now has `config/`, `config-overrides/`, `defaultconfigs/`, and `kubejs/` in `overrides/` (like how the `client.zip` was/is, unlike how `dist/client/` was).
    2. The above, but for `dist/dev/`
6. Removed the mode switch in the `build_pr` GH actions workflow, as it is now handled in the build script.

If any of these changes are not okay, let me know and I will adjust.

### DIFF
--- a/pack-mode-switcher.sh
+++ b/pack-mode-switcher.sh
@@ -55,29 +55,32 @@ if ! ([ -d "${NORMAL_CFG}" ] && [ -d "${HARDMODE_CFG}" ] && [ -d "${EXPERT_CFG}"
 fi
 
 if [ -z "$MODE" ]; then
-  if $SILENT; then
-    printf "\n${RED}Error: No mode specified in silent mode!${NORMAL}\n"
-    printf "\n${POWDER_BLUE}Accepted Inputs:\n${YELLOW}- [Normal, normal, N, n]\n- [Hard, hard, H, h]\n- [Expert, expert, E, e]${NORMAL}\n"
-    exit 1
-  fi
-
-  if [ ! -t 0 ]; then
-    printf "\n${RED}Error: Interactive mode entered in a non-interactive session!${NORMAL}\n"
-    exit 1
-  fi
-
-  # Interactive
-  CURRENT_MODE="$(cat .mode)"
-  CURRENT_MODE=${CURRENT_MODE:="N/A"}
+  CURRENT_MODE="$(head .mode)"
+  CURRENT_MODE=${CURRENT_MODE:="normal"}
 
   # Capitalise First Letter (only works in bash 4+)
   if ((BASH_VERSINFO[0] >= 4 )); then
     CURRENT_MODE=${CURRENT_MODE^}
   fi
 
-  printf "${POWDER_BLUE}Monifactory | Pack Mode Switcher${NORMAL}"
-  printf "\n${YELLOW}Current Mode: ${CURRENT_MODE}${NORMAL}\n\n"
-  printf "${POWDER_BLUE}Set Pack Mode:\nN: Normal    (The Default mode) \nH: Hard      (Adds more lines and progression, removes HNN and Monicoin spending) \nE: Expert    (A modifier for hard, enables some of the more extreme GTm settings among other things) \nSelection: [Normal / Hard / Expert]: "
+  if [ -t 0 ]; then
+    # Interactive stdin
+    printf "${POWDER_BLUE}Monifactory | Pack Mode Switcher${NORMAL}"
+    printf "\n\n${YELLOW}Current Mode: ${CURRENT_MODE}${NORMAL}\n"
+    printf "${POWDER_BLUE}Set Pack Mode:\nN: Normal    (The Default mode) \nH: Hardmode  (Adds more lines and progression, removes HNN and Monicoin spending) \nE: Expert    (A modifier for Hardmode, enables some of the more extreme GTm settings among other things) \nSelection: [Normal / Hardmode / Expert]:"
+    read MODE
+  else
+    # Non interactive stdin
+    echo $CURRENT_MODE
+    exit 0
+  fi
+
+  if ! $SILENT; then
+    printf "${POWDER_BLUE}Monifactory | Pack Mode Switcher${NORMAL}"
+    printf "\n${YELLOW}Current Mode: ${CURRENT_MODE}${NORMAL}\n\n"
+    printf "${POWDER_BLUE}Set Pack Mode:\nN: Normal    (The Default mode) \nH: Hard      (Adds more lines and progression, removes HNN and Monicoin spending) \nE: Expert    (A modifier for hard, enables some of the more extreme GTm settings among other things) \nSelection: [Normal / Hard / Expert]: "
+  fi
+
   read MODE
 fi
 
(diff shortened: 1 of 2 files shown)


pack-mode-switcher.sh
```diff
@@ -23,74 +54,62 @@ if ! ([ -d "${NORMAL_CFG}" ] && [ -d "${HARDMODE_CFG}" ] && [ -d "${EXPERT_CFG}"
...
 
-if [ -z "$1" ]; then
-  CURRENT_MODE="$(head .mode)"
-  CURRENT_MODE=${CURRENT_MODE:="normal"}
+if [ -z "$MODE" ]; then
+  if $SILENT; then
+    printf "\n${RED}Error: No mode specified in silent mode!${NORMAL}\n"
+    printf "\n${POWDER_BLUE}Accepted Inputs:\n${YELLOW}- [Normal, normal, N, n]\n- [Hard, hard, H, h]\n- [Expert, expert, E, e]${NORMAL}\n"
+    exit 1
+  fi
```
iluha168: This has no reason to exit, nor to *`printf` to 1 in silent mode*

I believe silent mode should still `read`, just without printing the help message

pack-mode-switcher.sh
```diff
@@ -23,74 +54,62 @@ if ! ([ -d "${NORMAL_CFG}" ] && [ -d "${HARDMODE_CFG}" ] && [ -d "${EXPERT_CFG}"
...
+
+  if [ ! -t 0 ]; then
+    printf "\n${RED}Error: Interactive mode entered in a non-interactive session!${NORMAL}\n"
+    exit 1
+  fi
+
+  # Interactive
+  CURRENT_MODE="$(cat .mode)"
+  CURRENT_MODE=${CURRENT_MODE:="N/A"}
+
```
iluha168: Normal mode is the default shipped mode, no need for N/A

There were no reasons to change these 2 lines

pack-mode-switcher.sh
```diff
@@ -23,74 +54,62 @@ if ! ([ -d "${NORMAL_CFG}" ] && [ -d "${HARDMODE_CFG}" ] && [ -d "${EXPERT_CFG}"
...
+  if $SILENT; then
+    printf "\n${RED}Error: No mode specified in silent mode!${NORMAL}\n"
+    printf "\n${POWDER_BLUE}Accepted Inputs:\n${YELLOW}- [Normal, normal, N, n]\n- [Hard, hard, H, h]\n- [Expert, expert, E, e]${NORMAL}\n"
+    exit 1
+  fi
+
+  if [ ! -t 0 ]; then
+    printf "\n${RED}Error: Interactive mode entered in a non-interactive session!${NORMAL}\n"
+    exit 1
+  fi
```
iluha168: Breaks BC in a big way - users are no longer able to get current mode, keep this check as was

pack-mode-switcher.sh
```diff
@@ -23,74 +54,62 @@ if ! ([ -d "${NORMAL_CFG}" ] && [ -d "${HARDMODE_CFG}" ] && [ -d "${EXPERT_CFG}"
...
+  printf "${POWDER_BLUE}Monifactory | Pack Mode Switcher${NORMAL}"
+  printf "\n${YELLOW}Current Mode: ${CURRENT_MODE}${NORMAL}\n\n"
+  printf "${POWDER_BLUE}Set Pack Mode:\nN: Normal    (The Default mode) \nH: Hard      (Adds more lines and progression, removes HNN and Monicoin spending) \nE: Expert    (A modifier for hard, enables some of the more extreme GTm settings among other things) \nSelection: [Normal / Hard / Expert]: "
+  read MODE
 fi
 
-function move_server_properties () {
-  # Only copy server.properties if it exists.
-  mv "$TARGET/server.properties" ./ 2> /dev/null || true
-}
```
iluha168, line 52: This is quite interesting, why did we have this in the first place? iirc i added this as a bug fix, but i dont remember why

tools/build/index.js
```diff
@@ -75,27 +75,40 @@ const cpSyncFiltered = (ourDir, newDir, filter) => {
...
+                `-Path "${resolve(`dist\\${group}\\*`)}"`,
                 `-DestinationPath "${resolve(`dist\\${group}.zip`)}"`,
             ])
             return;
         }
 
+        // Switch build to NM if mode is unset
+        if (!fs.existsSync("config/packmode.json")) {
+            const packSwitchPath = `dist/${group}${group !== "server" ? "/overrides" : ""}`; // to account for server build not being in overrides
+            await Juke.exec("chmod", ["+x", "./pack-mode-switcher.sh"]);
```
iluha168: git actually keeps the executable flag

so yeah you do not need this

tools/build/index.js
```diff
@@ -309,24 +315,24 @@ export const BuildDevTarget = new Juke.Target({
...
         fs.mkdirSync("dist/.devtmp", { recursive: true });
         for (const folders of includeList.filter(v => !(v === "mods" || v === "config"))) {
-            symlinkSync(resolve(folders), resolve(`dist/dev/${folders}`));
+            symlinkSync(resolve(folders), resolve(`dist/dev/overrides/${folders}`));
         }
 
         // "merge" both mod folders
         fs.cpSync("dist/modcache", "dist/.devtmp", { recursive: true });
         fs.cpSync("mods", "dist/.devtmp", { recursive: true, force: true });
         symlinkSync(resolve("dist/.devtmp"), resolve("dist/dev/mods"));
```
iluha168: missed dist/dev/mods?

iluha168 (approved):
Honestly quite incredible

pack-mode-switcher.sh
DragonDuck24: No clue. I checked the git history and we never had any server.properties in the different modes.

tools/build/index.js
DragonDuck24: oh sick i didnt realize

pack-mode-switcher.sh
DragonDuck24: I was under the assumption I should still print errors in silent mode, and only truly be silent if it succeeds
But yes I can have it still read and not print anything in silent mode

pack-mode-switcher.sh
DragonDuck24: Will revert

pack-mode-switcher.sh
DragonDuck24: Okay yeah i have no idea why i did that. I have fixed it

tools/build/index.js
DragonDuck24: I didnt think mods were supposed to go in overrides. Are they?

DragonDuck24: It has been moved in

PR #2039, opened 2025-08-23
Adds a python pack switcher

Apoptosis847639:
This removes the old batch and shell pack switchers in favor of a python version


Apoptosis847639:
first and last one Dragon is doing,
I'll update the 2nd one

DragonDuck24:
You should probably be doing those changes. In the time between our PRs being merged, the GH action workflow will be borked, and same with the build script.
I will then change it again to reflect my changes. You made this change, and should keep the repo working without mine, as our changes are (mostly) independent.

Apoptosis847639:
> You should probably be doing those changes. In the time between our PRs being merged, the GH action workflow will be borked, and same with the build script. I will then change it again to reflect my changes. You made this change, and should keep the repo working without mine, as our changes are (mostly) independent.

ok i'll do that

WithersChat:
Can we not? I don't like python.

Apoptosis847639:
> Can we not? I don't like python.

Why?

pack-mode-switcher.py
xilni, line 1: This file could really do with a pass by a proper linter like [ruff](https://github.com/astral-sh/ruff) and bonus points if you round out the type hinting with a mypy or even better a ty pass for sanity.

WithersChat:
We will not.

PR #2040, opened 2025-08-24
Pterb Retier Fix

PlasmaticVoide:
Pterb had its research recipe still, and because the gt recipe wasn't properly removed, the soldering alloy script was creating new recipes based off the more expensive recipe.

### DIFF
--- a/kubejs/server_scripts/fixes_tweaks/pterb_retier.js
+++ b/kubejs/server_scripts/fixes_tweaks/pterb_retier.js
@@ -3,8 +3,8 @@
  * - Replaces UV components with LuV components.
  */
 ServerEvents.recipes(event => {
-    event.remove({ id: "gtceu:pterb" })
-    event.remove({ id: "gtceu:research_station/1_x_gtceu_active_transformer" })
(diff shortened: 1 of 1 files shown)


madeofstars0:
Is the pterbEnabled config value (gtmutils.yaml) related to what you are trying to accomplish here?

https://github.com/ThePansmith/Monifactory/blob/main/config/gtmutils.yaml#L19

It is set as true in both the default and the normal configs (I didn't check the others).

PlasmaticVoide:
the pterbEnabled config in gtmutils.yaml is a config that either enables or disables the creation of the pterb multiblock
In moni, we change the pterb recipe to tier it at a different point in progression, using the pterb_retier.js file. This file however had some mistakes in it that meant it didn't fully properly remove the old assline and research recipes, which is what this pr addresses.

madeofstars0:
I see, I just saw your PR and was mucking thru the configs looking for something else and saw that entry. Thanks for the explanation, there is wayyyyyyyyyy too much to know.

Xefyr0 (approved):

PR #2041, opened 2025-08-24
Gives a GUI to the packmode switcher in monilabs

Apoptosis847639:
Do not merge this until my PR in MoniLabs gets merged and this PR will have a commit updating monilabs.


Apoptosis847639:
even with the new MoniLabs I want to work on giving the UI a background and overall refining it so yeah this is staying as a draft.

PR #2043, opened 2025-08-24
Better texture for Liquid Sculk.

Flare32:
also has a texture for hadal sculk

i just want to be done with these right now


Xefyr0 (changes requested):
As I've mentioned earlier, these fluids are definitely brighter overall compared to the current textures.
The color palette you used is highly similar and that's good, but on the website the "files changed" tab does a side-by-side comparison and it's clearly visible that the new fluids have significantly more highlights than the block textures. I think the highlights on Hadal Sculk are also brighter in and of themselves (being pure white instead of a light gray)

I don't think these fluid textures have to perfectly match in highlight amount (a little increase is good actually) but it should be closer than things currently are.

WithersChat:
Agreed with Xef on that one, but only for hadal sculk. Should make the 2 more distinct too.

PR #2044, opened 2025-08-24
Fix moving coremod materials

PlasmaticVoide:
When bioalloy and cmatrix were moved to coremod, the double plates and long rods and small gears were left behind
Also fixed c-matrix mesh smelting having conflicting ebf recipes


Xefyr0 (changes requested):
Looks like you accidentally nicked Rhodium-Plated-Palladium in kubejs/server_scripts/microverse/miners.js

PR #2046, opened 2025-08-25
Nuke Ultradense Hydrogen and Ultradense Helium.

Flare32:
### Does as title says; removes Ultradense H/He because they have one use, bad textures, and no reason to exist
*Flarenuke, verb: "To remove an item to avoid having to update it's texture."*

~~...Yes, yes I *did* forget to remove the textures.~~
**Okay so I don't think I *can* remove files just using the web UI, so they'll probably just have to sit around until someone who knows how to use anything other than the website can clear them out. :(**


WithersChat:
Why 2 slots of oxygen plasma? But aside from that YES THANK YOU

PR #2055, opened 2025-08-30
fix: Pack mode switcher bug when target folder has spaces in it

iluha168:
This pull definitely does not contain a harmless refactor inside


Xefyr0 (approved):

iluha168:
GitHub what the f

PR #2068, opened 2025-09-05
Revamped Rare Earth Line

Jollahs:
makes rare earth outputs better thumbs up emoji


Xefyr0 (changes requested):
Actually, could you please combine the rare earth + terbium files into a single file since it's only one processing line?

WithersChat:
So using the new RE line gives more of all the old stuff + some terbium, right?

PR #2084, opened 2025-09-09
Buff naqfuelline

Xefyr0:
Also improves some aspects of late-game radioactive isotope management. 
Abyssal Cores use Americium instead of Californium (cheaper)
and Activated Netherite production uses half as much Berkelium instead.

At Jollah's behest:
https://discord.com/channels/914926812948234260/1229929078547550238/1414759327570464768


Jollahs:
Californium is fine in cores, you dont need as many. and stabilizing it is 144mb to 8 stabilized. It was the liquid you needed. It made it so that 4 pieces of californium were required for each netherite piece, which was overkill.

Issue #2067, opened 2025-09-04
[Bug]: Steam Turbines from Thermal Systeams are broken


esmenard:
### Issue Summary

Unupgraded Steam Turbines from Thermal Systeams only produce 20 RF/tick in the 0.13 test build, which makes them unusable past the very early game and breaks both 1 boiler + 2 turbines being better than 3 dynamos and the 1 boiler for 2 turbines ratio.

When I start a new world, the config file that is generated in `saves\New World\serverconfig\systeams-server.toml` is different from the one in `defaultconfigs\systeams-server.toml` (https://github.com/ThePansmith/Monifactory/blob/c8182a87982e04ac820a1630163969195af21cb4/defaultconfigs/systeams-server.toml) that was installed with the pack. The config file that I get instead is the following : 
```#If the Dynamo in dynamo augment tooltips should be replaced with Dynamo & Boiler
#This doesn't have as much support for translations (it can still be translated with the key info.systeams.augment.type.DynamoBoiler)
replace_dynamo_augment_tooltips = true
#Systeams Config
#---------------
#The multiplier on the steam dynamo's RF/t
#Range: 10 ~ 32000
steam_dynamo_base_power = 20

["Steam Values"]
	#The number of mb of steam produced per RF of energy usually produced by the same fuel in a dynamo
	#Note that this does not affect the steam dynamo's rates. That needs to be adjusted with a datapack
	#Range: 0.05 ~ 100.0
	stirling = 0.5
	#Range: 0.05 ~ 100.0
	magmatic = 0.5
	#Range: 0.05 ~ 100.0
	compression = 0.5
	#Range: 0.05 ~ 100.0
	numismatic = 0.5
	#Range: 0.05 ~ 100.0
	lapidary = 0.5
	#Range: 0.05 ~ 100.0
	disenchantment = 0.5
	#Range: 0.05 ~ 100.0
	gourmand = 0.5
	#Range: 0.05 ~ 100.0
	pneumatic = 0.5
	#Range: 0.05 ~ 100.0
	frost = 0.5

["Boiler Speed Multipliers"]
	#The speed multiplier on each boiler's mB/t
	#Range: 0.05 ~ 20.0
	stirling = 16.0
	#Range: 0.05 ~ 20.0
	magmatic = 16.0
	#Range: 0.05 ~ 20.0
	compression = 16.0
	#Range: 0.05 ~ 20.0
	numismatic = 16.0
	#Range: 0.05 ~ 20.0
[...]

PR #2085, opened 2025-09-09
Update systeams config

DragonDuck24:
Updates the systeams config to match the 1.9 update.
Makes turbines output the proper amount of power (160rf/t)

At some point we could revisit water to steam ratios with a mixin, but I'm not worrying about that right now.

### DIFF
--- a/defaultconfigs/systeams-server.toml
+++ b/defaultconfigs/systeams-server.toml
@@ -9,23 +9,23 @@ steam_dynamo_base_power = 160
 	#The number of mb of steam produced per RF of energy usually produced by the same fuel in a dynamo
 	#Note that this does not affect the steam dynamo's rates. That needs to be adjusted with a datapack
 	#Range: 0.05 ~ 100.0
-	stirling = 0.5
+	stirling = 2.0
 	#Range: 0.05 ~ 100.0
-	magmatic = 0.5
+	magmatic = 2.0
 	#Range: 0.05 ~ 100.0
-	compression = 0.5
+	compression = 2.0
 	#Range: 0.05 ~ 100.0
-	numismatic = 0.5
+	numismatic = 2.0
 	#Range: 0.05 ~ 100.0
-	lapidary = 0.5
+	lapidary = 2.0
 	#Range: 0.05 ~ 100.0
-	disenchantment = 0.5
+	disenchantment = 2.0
 	#Range: 0.05 ~ 100.0
-	gourmand = 0.5
+	gourmand = 2.0
 	#Range: 0.05 ~ 100.0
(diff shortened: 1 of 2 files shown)


DragonDuck24:
addresses #2067

defaultconfigs/systeams-server.toml
```diff
@@ -1,41 +1,31 @@
...
-	gourmand = 0.5
+	gourmand = 2.0
 	#Range: 0.05 ~ 100.0
-	pneumatic = 0.5
+	pneumatic = 2.0
 	#Range: 0.05 ~ 100.0
-	frost = 0.5
+	frost = 2.0
 
 ["Boiler Speed Multipliers"]
```
Xefyr0, line 30: Nearly there.
When I tested this PR in-world, a stirling boiler could feed 8 Steam Dynamos instead of the expected 2.
This makes sense - when we quadrupled the boiler's steam value to make it 4x as efficient as the corresponding dynamo, it also made the boiler 4x faster.

Thus, the boilers' speed multipliers should be quartered to counteract this.

DragonDuck24: I thought I checked that it was the same whoops. I was eepy and maybe i didnt
I'll update it soon

DragonDuck24: Okay I made this change, however I am noticing that my steam turbines are using 1/10 the steam required. The recipe is 100mb -> 2000rf, a 1:20 ratio instead of the 1:2.
So as it stands we have 1mb water -> 160mb steam in boiler, then 160mb steam -> 3200rf.
Do we need a to mixin the turbine recipe?

DragonDuck24:
Drafting for now, I'll look into it again when I get home later

defaultconfigs/systeams-server.toml
Xefyr0: That's weird. When I tested it in-world as I was talking about above, 1 boiler at 640 mB/t struggled to fuel 8 turbines, so each turbine would only have been getting 80 mb/t tops. This dropped down to 160 mB/t after the boiler speed reduction, which fuels 2 turbines at 80mB/t each.

Moreover, putting down a new turbine and giving it a bucket of GT steam produces 2080 FE; 2 FE per mB steam as it should be (Plus 80 FE extra for some reason, but that's negligible)

DragonDuck24: Testing again making sure everything was correct, I am now getting the desired results with the current state of the PR. Feel free to test again but to me it looks good now, and will undraft.

Xefyr0 (approved):

PR #2094, opened 2025-09-09
Update creative chest recipe to reflect new content & items

Xefyr0:


WithersChat:
As I said, could use a final rework later down the line but this is infinitely better than what we got now so it's good for me.

PR #2095, opened 2025-09-09
Add more snowflake activity to Cryotheum Dust

Xefyr0:
Thought the snowflakes in the cryotheum dust texture looked a little sparse.
This adds 2 more into the cycle.


WithersChat:
No. Would clash too much with primal mana which already has its snowflake rep fighting for its life due to space constraints.

PR #2110, opened 2025-09-13
Make TES not consume infinity ingots like Joey Chestnut

Apoptosis847639:
It does this by
A) Reducing the amount of infinity dust the recipe consumes by 1/3, 3 tiny dust
B) Decrease the amount of Naq. SuperFuel the TES consumes per second to 325 MB from 500 MB

This approx. consumes 1 infinity ingot every 5 minutes(if my math isn't bad lol), 5x more efficient than before


Xefyr0 (approved):

WithersChat:
I'd even suggest bringing infinity down to 1 or 2 tiny dusts. Or 1 small dust.

Apoptosis847639:
> I'd even suggest bringing infinity down to 1 or 2 tiny dusts. Or 1 small dust.

I'll do 2 tiny dusts.

PR #2120, opened 2025-09-18
Add faulty memory card recipe

mosesyu1028:
The faulty memory card from MAE2 currently doesn't have a recipe, so I added a new recipe taking a regular memory card and crafting with an ae2 knife. Please feel free to edit the recipe as needed.

Maybe we could mention the existance of it (and how to use decrement mode) in the QB for assembly line?

image

---

Also, the advanced memory card from BetterP2P seems to have a duplicated recipe and the extra recipe isn't actually needed:

Current build:
image

Changed build:
image

### DIFF
--- a/kubejs/server_scripts/mods/AE2.js
+++ b/kubejs/server_scripts/mods/AE2.js
@@ -1099,8 +1099,8 @@ ServerEvents.recipes(event => {
         .duration(200)
         .EUt(GTValues.VA[GTValues.MV])
 
-    // BetterP2P
-    event.shapeless("betterp2p:advanced_memory_card", ["ae2:memory_card", "ae2:network_tool"])
+    // Faulty Memory Card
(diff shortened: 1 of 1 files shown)


Apoptosis847639:
and it seems I am an idiot and didn't read the rest of it :sob:

Apoptosis847639:
After looking at it again, I can't seem to find the other recipe, might be better to remove that one and keep the recipe in AE2.js

mosesyu1028:
> After looking at it again, I can't seem to find the other recipe, might be better to remove that one and keep the recipe in AE2.js

What do you mean by the other recipe? I could keep the betterp2p recipe in for redundancy if maybe my version of the pack is broken or something.

Apoptosis847639:
> > After looking at it again, I can't seem to find the other recipe, might be better to remove that one and keep the recipe in AE2.js
> 
> What do you mean by the other recipe? I could keep the betterp2p recipe in for redundancy if maybe my version of the pack is broken or something.
Nevermind, I think it's fine

PR #2123, opened 2025-09-19
Antimatter stuff downtier + protomatter buff

PlasmaticVoide:
Buffs protomatter by a factor of many, in response to jollahs' feedback
Downtiers all the antimatter stuff to UV, making it immediately competitive with large naq and naqfuel (by withers' request)


Xefyr0 (changes requested):
Actually, can you move the antimatter quest in the Power Generation chapter to match its new tier? Putting it 25 or 50% of the way from UV to UHV probably reflects the balancing of this PR the best.

PR #2125, opened 2025-09-19
Platine and Indium related ore tags. 

Phoenixvine32908:
This PR seeks to add tags called gtceu:purified_platline_ores and gtceu:purified_indium_ores to help make ore processing systems easier with tag export/storage busses.


Phoenixvine32908:
Got a bit tired of manually writing out all the ores in the tag export/storage bus so I thought this was a good solution.

Xefyr0 (changes requested):
Argentite can also be used for Indium, so make sure to include it too.

PR #2126, opened 2025-09-20
Transcendental matrix line

Jollahs:
still needs texture(s)
needs to be playtested (probably by me)
not all item types are final


WithersChat (approved):

Xefyr0:
I believe this line may need further review - it is not clear to me what makes this line enjoyable & unique especially after we identified that the endgame has an issue with there being too many processing lines.

That will have to be done as a group, and it will be helpful for us all to experience the gameplay the line brings to the table in the dev run so I am merging this.

PR #2128, opened 2025-09-20
Naqfuel/Curium rebalance

Xefyr0:
Changes:
2.67x as much naquadah isotope yield (and by extension Curium) per Naquadria/Naquadah dust in liquid naqfuelline.
Curium cost for microminers slashed
Berkelium cost for Sculk Catalyst duplication, Actneth slashed
Increased Plutonium yield from T8.5MM
Buffed Quadium plasma to make it less useless
Solid naqfuelline (The one that makes rods) now yields small amounts of Curium in place of Plutonium.

tl;dr: Curium, Berkelium, Californium are easier to get but you also get a _lot_ more out of the same amount.
Solid naqfuelline costs 468 Enriched Naquadah and 117 Naquadria to get 10 Curium dust, and liquid naqfuelline costs 20 Naquadah and 6 Naquadria to get roughly 10 dusts' worth of liquid Curium.

Honestly I worry that liquid naqfuelline is overtuned in terms of Curium production now, but that's probably what y'all want :V


WithersChat:
We'll test the viability of this by trying to beat the pack without any new NCN reactor.

PR #2130, opened 2025-09-21
Texture Updates: Round III

88Flare:
### Contains a MULTITUDE of texture changes.
*121 files changed?????? One-hundred and twenty-one???????????? And they're all new ones?????????????????????*

Includes:
- Energy Cores, all three of them
- Crystal Matrix Line
- Mob parts and related items (no warden horns since those were already good)
- Thrusters
- Solidified & Stabilized Elements (stabilizied ones still need the animation to be toned down btw)
- Almost *ALL* NCN Isotopes and Fuels

No images this time since uh, that would be a lot of images.


WithersChat:
Might need to redo the thrusters but that's 1.0 material

PR #2134, opened 2025-09-22
fixed dark steel casing path in the multiblock_registry

Saring:


Xefyr0:
Another fix for this issue has already been implemented in #2130.

PR #2143, opened 2025-09-25
Nerf re line outputs

Jollahs:
devrun should NOT have 500k europium dust from re line so i decided to dial it all back a bit


Jollahs:
also fixes plasma conflict in hyperbolic projector

PR #2152, opened 2025-09-28
Tweak lower tier microminer missions

Xefyr0:
Shuffles some missions around so that there are fewer useless microminer missions.
Also groups like with like a bit more, except for platinum group sludge ores which are spread out instead to give players a challenge.
Lastly, some values were tweaked. (Gold yield in particular was drastically reduced)


WithersChat:
...please don't split ores between resource types. Now THAT would be tedious. It's already tricky enough to find a good level emitter balance without making it worse.

PR #2155, opened 2025-09-28
UHV and UEV voltage coils

PlasmaticVoide:
Adds voltage coils for UHV and UEV tiers, made of darconite and Ns respectively. 
This notably now gates these energy hatches behind the singleblock assemblers of those tiers.
Also changes the textures of the ZPM and LuV voltage coils to match the materials they're made of. 
This pr is missing the textures of UEV coil because I can't recreate the necrosiderite effect
Reason for moving the energy hatches outside the `converters` loop was because all three tiers have more changes than was expressed in the parameters of `converters`


Xefyr0 (approved):

Xefyr0:
Would prefer that the "4 pixels" be kept & used on the UHV and UEV coils so that we make only minimal texture alterations to not screw with players' resource packs, but this is passable.

PR #2157, opened 2025-09-28
Remove spark due to freezing linux systems

Apoptosis847639:
This is due to spark freezing systems on the latest Linux kernel release.


Apoptosis847639:
Btw spark can be readded when the bug is fixed, this affects every linux user running the latest linux kernel version.

PR #2158, opened 2025-09-29
Give Null mass

Jollahs:
dont divide by zero kids
gives null a mass of 1 by adding a proton


Apoptosis847639:
just curious did this break something or no?

Xefyr0 (approved):

Xefyr0:
Null is secretly Hydrogen confirmed

WithersChat:
Sad that this is needed but fair enough.

PR #2150, opened 2025-09-27
Misc fixes

WithersChat:
- Just implementing agreed-upon changes
- That's it, that's the PR.

PR #2163, opened 2025-10-03
Superfuel tweaks

Xefyr0:
Change Eltz -> Actinium in superfuel recipe, and slight nerf to Superfuel's power generation.

This doubles the LNRs needed for beating the pack (According to Jollahs) from 2 to 4, but only decreases fuel energy density by 16.6%.


WithersChat:
Integrated into #2150 because the TES recipe change and also Jollahs recommended a bit of a further nerf of superfuel (40mb per recipe instead of 30)

PR #2174, opened 2025-10-11
transgebderline but i got the branch right this time :sobbign:

Flare32:
oops


Flare32:
xp juice is alr used in droves for sculk stuff and it's kinda like dirt cheap

PR #2175, opened 2025-10-11
Split off ABS step in biosolder and bioalloy 

Xefyr0:
This way, Bioalloy and Biosolder use ABS for an "alloy base" made by mixing metal dusts, the output of which is used in SVat for the finishing step.


Xefyr0:
Added XP requirement to Bioalloy.
Existing ones were tweaked to widen their range & lower midpoints so that the Bioalloy requirement could fit in as the highest & narrowest.
Backwards-compatible with existing XP hatch solutions.

PR #2191, opened 2025-10-16
Pack-mode-switcher.sh changes

zxro:
I had an issue of menu message printing twice on my artix linux system (wezterm terminal).
Also, script crashed if was launched with no args when no .mode file exists.

I did some minor readability fixes (mostly to printf lines).
Line 62 changed () to {} for optimization.

Changed logic to never print or request anything twice.
Tested on my machine, works fine.

### DIFF
--- a/pack-mode-switcher.sh
+++ b/pack-mode-switcher.sh
@@ -1,6 +1,5 @@
 #!/bin/bash
 
-
 # Colors
 RED=$(tput setaf 1)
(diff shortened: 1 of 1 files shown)


Apoptosis847639:
We already have plans to remove the shell and bat files in favor of Monilabs(The coremod) doing it.

Apoptosis847639:
after asking the other contributors since 0.12.9 people still exist, this *might* get merged

PR #2193, opened 2025-10-17
Nerf JEAN & transform up to shorten burntime

Xefyr0:
Unfortunately prevents JEAN from working in LV generators, but will help prevent generators from operating at reduced parallels for long past the period of limited demand.


Xefyr0:
Merged at Withers' approval in Discord

PR #2194, opened 2025-10-18
fixed server installation section

Brandon4466:
Steps were missing to move files from overrides folder to main server folder. fixed this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,7 @@ To spice up your Monifactory experience, you can add any of the following mods t
 4. Move the server.zip and forge installer into your server directory. The rest of the guide assumes your current directory is the server directory.
 5. Run the forge installer and install the forge server, this can be done with the command ``java -jar TheForgeInstallerName.jar --installServer``
 6. ``unzip server.zip``
+7. Move the contents of the overrides folder (from server.zip) into the server directory, this can be done with the command ``mv overrides/* .``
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +107,7 @@ To spice up your Monifactory experience, you can add any of the following mods t
 4. Move the server.zip and forge installer into your server directory. The rest of the guide assumes your current directory is the server directory.
 5. Run the forge installer and install the forge server, this can be done with the command ``java -jar TheForgeInstallerName.jar --installServer``
 6. ``unzip server.zip``
+7. Move the contents of the overrides folder (from server.zip) into the server directory, this can be done with the command ``mv overrides/* .``
 7. Use ``./run.sh`` to generate the eula.txt, then again after you accepted run it again to start the server. Modifying the server.properties file to change the port may be neccesary.
```
Apoptosis847639, line 111: Both this and the next line should be renumbered to be 8 & 9.

DragonDuck24:
These steps used to exist, but we're removed with a change to how server files are built in (#2038)
The current steps are correct for 0.13, but not for 0.12.
Lead devs, should we revert and readd these steps until 0.13 is released, or leave as is?

Xefyr0:
Ideally, we should provide steps for both versions.

Xefyr0:
Resolved with #2227

PR #2197, opened 2025-10-18
Bioalloy Base Changes

PlasmaticVoide:
- Removed the dust form of Bioalloy Base because it existing only gave the alloy a mixer recipe, bypassing the Omnic Matrix Coil gate of Bioalloy
- Gave the ABS recipes of Bioalloy Base program circuits to properly distinguish between the two recipes
- Removed decomposition by centrifuging because that recipe consumed 1B of bioalloy base and gave back the materials to craft 1,008mB of bioalloy base, meaning that performing 125 recipes would magically grant you 126 recipes worth of materials back (lol)


Xefyr0 (approved):

WithersChat:
Bioalloy base is still a bit boring but that's 1.0 matters.

PR #2203, opened 2025-10-19
swap adjacentFluid to accept custom fluids

PlasmaticVoide:
Fix provided by Onion
Fixes netherrack and grains of infinity recipes
Only needed for those two recipes, but changed all instances to avoid confusion of having two different methods.


Apoptosis847639:
mb yall :pensive:

PR #2204, opened 2025-10-19
Make fluid condition order for rock crusher consistent

Xefyr0:
In EMI, the base GTM rock crusher recipes have lava first, and ours have it second. This makes the order consistent.


WithersChat:
This will conflict with #2203 to hell and back probably.

PR #2209, opened 2025-10-19
Change HDD, Sculk Supercon recipes to match QB

Xefyr0:
Gets rid of all forms of HDD and Sculk Superconductor except wires (and fine wire + foil for HDD) and adds recipes so that the forms that *do* exist can be crafted in the EBF/Vac Freezer.


WithersChat:
Good on removing the unused forms, but why a recipe for every form? Why not just single wire and foil?

Plus, if you're dipping it in something, you need more surface area than a hex wire has.

PR #2230, opened 2025-10-26
Update monilabs to 0.18.2 and gregtech to 7.2.1

NegaNote:


Xefyr0 (changes requested):
GTM update relies on a JEI mod update and causes KJS errors :skull:
Either those should be fixed or the update removed from this PR.

Xefyr0 (approved):

PR #2232, opened 2025-10-27
The entire fucking chest rework

WithersChat:
- Yes, that's a single commit
- Also includes textures for new material forms
- The Monium mission also got slightly buffed


kubejs/assets/gtceu/models/item/material_sets/infinity/foil.json
```diff
@@ -0,0 +1,6 @@
+{
+  "parent": "item/handheld",
```
NegaNote, line 2: should be item/generated

WithersChat: All of infinity is like that 😭

NegaNote: :thevoices:
is infinity stuff *intended* to be held like tools? if not yeah it should all be made to be item/generated

Flare32: so i tested this a while back after i think you pointed it out? and it seems like item/handheld and item/generated act functionally the same

Flare32: Screenshot from 2025-10-26 18-03-07

NegaNote: did you look in third person? that's where the difference is

Flare32: i did not but i do not think it will matter because if i missed it while actively trying to look for the differences it will not be encountered in normal play i don't think

NegaNote: i guess so

PR #2252, opened 2025-11-01
Remove Difficulty Lock and replace functionality with Default Options

Xefyr0:
It turns out that we have two mods that do the same thing - both Difficulty Lock and Default Options allow for us to configure a default, locked difficulty at the start of every world creation.

This PR gets rid of the Difficulty Lock mod (made by Serilium) and migrates the functionality over to Default options (made by BlayTheNinth) to eliminate a superfluous mod.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -416,11 +416,6 @@
       "required": true
     },
     {
-      "fileID": 6429218,
-      "projectID": 342584,
-      "required": true
-    },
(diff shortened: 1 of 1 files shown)


Xefyr0:
It turns out that Default Options does not properly lock the difficulty like Difficulty Lock does. Closing this PR.

PR #2256, opened 2025-11-02
Powergen chapter fixes

Xefyr0:
Remove UIV from the voltage scale & a bunch of minor tweaks to improve quest layout and descriptions.


Flare32:
UIV might still want to be on the timeline since like   UIV wires exist and such

Xefyr0:
> UIV might still want to be on the timeline since like UIV wires exist and such

This is for power generation, not for power transportation.
Power gen options are placed on the timeline based on a combination of where you unlock them in progression and their base generation amount, and nothing is unlocked in UIV or can make use of a UIV output hatch since those are uncraftable.

PR #2262, opened 2025-11-04
update monilabs to 0.18.8

NegaNote:
* add new sound for microverse projectors courtesy of Phoenixvine
* allow microverse projectors to keep their microverse type when the structure is broken, if the microverse does not passively decay
* allow omnium synth progress the first time an element is input
* prevent :null: in the creative energy multi render when shaders are loaded

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1081,7 +1081,7 @@
       "required": true
     },
     {
-      "fileID": 7187029,
+      "fileID": 7190843,
       "projectID": 1266569,
       "required": true
     },


WithersChat:
Need to fix the ORB first 😭

NegaNote:
Moni Labs 0.18.8

* Fix ORB not working across dimensions (we don't talk about 0.18.7)
* Slight fix for hyperbolic projector tooltip
* Fix SVat minimum casings and make the laminated glass mandatory

PR #2267, opened 2025-11-06
update monilabs to 0.18.10

NegaNote:
Allow higher-tier glasses on Microverse Projectors Remove limits on custom sensor hatches per multi

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1081,7 +1081,7 @@
       "required": true
     },
     {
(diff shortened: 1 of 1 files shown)


NegaNote:
update to 0.18.10
* Change PrismaC structure to not allow hatches on its top
* Adjust minimum casing requirement to be slightly more lenient

Issue #2263, opened 2025-11-04
[Bug]: Silver Furnaces are slower than Copper despite being next in the furnace tiers


tyreforhyred:
### Issue Summary

Silver furnaces are stated to take 80 ticks to smelt an item in the quest book, however the shipped config has it set to 140.

### Reproduction steps

observe the tooltip and smelting speed of a silver furnace

### Relevant log output

```shell

```

PR #2268, opened 2025-11-06
Various Questbook Fixes

### LINKED ISSUE #2266: [Bug]: The 'Gemstone Sensors' quest is locked with no way of completing other than forcing it with cheats
### Issue Summary

The quest 'Gemstone Sensors' with the dependency 'The First Tier Three Circuit' seems to be locked permanently without me being able to complete it, even though I have made an advanced integrated circuit for the dependent quest a while ago (I am in EV now). Making the gemstone sensor just puts a gray checkmark on it, indicating I have the materials to complete it, but it hasn't unlocked yet.

### Reproduction steps

1. Craft an advanced integrated circuit and complete the quest ('Gemstone Sensors' quest dependency)
2. The 'Gemstone Sensors' quest is still locked; crafting a gemstone sensor does not complete it.

### Relevant log output

```shell

```

PlasmaticVoide:
ncn and platline changes put into qb
also other qb fixes that have been reported from builds
Scaling Up chapter not going to be left like that, just wanted to get an accurate version of qb out for next build. At some ponit i'll be redoing that chapter to look nice. 
silver furnace made to be better than what it's crafted from and worse than what it's crafted into
fixes #2266, #2263,


Sobek88:
#2266 also affects the blaze powder quest (and possibly the other elemental powder quests); I completed the Elemental Reduction Fluid and Neural Networking quests, but obtaining blaze powder does not complete the quest. Instead, I see a gray check mark.

PR #2279, opened 2025-11-11
Updating ae2 addons and removing ae2 itself in preparation for AE2C.

Phoenixvine32908:
Updates ae2wtlib, extended ae, and monilabs.
Removes ae2 since we now have monilabs with the bundled fork in jar.


NegaNote:
bruh this needs to be a branch, right now its just listing the itnt mission as the only change

Phoenixvine32908:
I'll redo the thing. mb

PR #2286, opened 2025-11-12
move from JEMI to TooManyRecipeViewers

NegaNote:
all JEI item/fluid hiding is now done via tags.


NegaNote:
Realized I'm not sure how to go about the specific whitelisting that we had in the system before this PR for things like the example AE2 facade and the NCN ferroboron ingot. I'm open to suggestions.

NegaNote:
Turns out addItems *does* work -- for everything except the example AE2 facade that was otherwise removed by the tags earlier. Again, taking suggestions.

WithersChat:
Drafted until post-0.13
Too many things using up mental bandwidth for now

WithersChat:
Closing since this is a bit of a forgotten thingy.

PR #2290, opened 2025-11-14
buff magma cream output from MMs for an alternative lava source

NegaNote:


Xefyr0:
The players yearn for the T4 Dense Magma Block mission \s

WithersChat:
Wanted to slightly reduce lava costs as well TBF but this works too?

PR #2298, opened 2025-11-18
Solder progression redo

Xefyr0:
Solder progression changes as described in #moni-dev.

Previously:
Adv. Solder **optional in EV,** required in ZPM.
Living Solder **optional in ZPM,** required in UEV.

With this PR:
Adv. Solder optional in HV, **required in EV.**
Living Solder optional in LuV, **required in ZPM.**


WithersChat:
Slight change request: I think that every recipe should allow at least 2 different solder tiers, even if the higher tier isn't unlocked yet. So living solder should be usable IV and up even if you can't make it yet.

Xefyr0:
> Slight change request: I think that every recipe should allow at least 2 different solder tiers, even if the higher tier isn't unlocked yet. So living solder should be usable IV and up even if you can't make it yet.

These recipes would be not be used. They are noob traps.
Remember: Advanced Solder is more expensive than regular Solder, and Living solder is comparable in cost to Advanced Solder. (AFAIK)
Players would not (and should not) go back and put in the effort to change their patterns to a more expensive solder material. to do so would be entirely detrimental.
Before this PR, doing the higher-tier solder on recipes that didn't strictly require it had the benefit of letting you skip Advanced Solder. After this PR it does not.

The only reason any recipes that allow for more than one solder tier to exist (EV and ZPM) is to smooth the transition you have to make at those tiers. Even then, I doubt that the more advanced option available will see use since it's not strictly necessary.

It also helps to reduce recipe bloat.

PR #2309, opened 2025-11-23
update gtm to 7.3.0

NegaNote:
* had to fix some recipes that are errors now but were silently failing before


WithersChat (approved):

Xefyr0 (approved):
Whoops. Not me forgetting that Vanadium Pentoxide doesn't exist in NM :skull:

PR #2312, opened 2025-11-23
fix weblate conflict hell

iluha168:


Xefyr0 (changes requested):
1. Lang files without any translated keys should be deleted.
2. Several lang files have incorrect locale codes: pt_BR -> pt_br, pt_PT -> pt_pt, nl -> nl_nl, and so on.

Xefyr0:
Clearly Git and Windows are colluding against me.
Someone else will have to fix this PR, I'm afraid

PR #2330, opened 2025-11-30
add germanium to the neptunium palladium aluminium alloy for iv hatches

NegaNote:
will result in the original alloy being yeeted


Xefyr0:
For Withers when she reviews this:
I think this PR is not necessary. There's already the ABS, Multiblock Chem Bath, Signalum, Advanced Projector, and T4MM between the two halves, and the quests are already separate. IMO This would be overdoing it.

If you agree, please close the PR.
If not, we should discuss further in #moni-dev.

PR #2349, opened 2025-12-07
Add Material Info to many recipes

### LINKED ISSUE #1354: [Bug]: Atomic Reconstructor is missing recycling recipes
### Issue Summary

There are no recycling recipes (Plasma Scrapping in Arc Furnace, Part Grinding in Macerator, etc) for any of the Atomic Reconstructors.

### Reproduction steps

Attempt to recycle any Atomic Reconstructor in an Arc Furnace.
Also no recipe in EMI.

### Relevant log output

_No response_

Xefyr0:
Resolves #1354 and #2411.

Unfortunately, this PR creates as many bugs as it resolves - many recycling recipes don't recognize the material info of one or more ingredients due to load order.

The solution for this looks to be the same as Screret describes [here](github.com/GregTechCEu/GregTech-Modern/issues/3295)

Unfortunately, this would require a major rework of how our custom recipes would be structured and in some cases is simply untenable. Imagine dealing with things like the Causality Collapse Device, which requires require 8 recursive steps and in turn 8 separate files!

### DIFF
--- a/kubejs/server_scripts/gregtech/energy_cores.js
+++ b/kubejs/server_scripts/gregtech/energy_cores.js
@@ -44,21 +44,21 @@ ServerEvents.recipes(event => {
         .itemOutputs("kubejs:empty_bathyal_energy_core")
         .duration(10)
         .EUt(GTValues.VA[GTValues.IV])
-        .addMaterialInfo(true)
+        .addMaterialInfo(true, true)
 
     event.recipes.gtceu.discharger("kubejs:empty_abyssal_energy_core")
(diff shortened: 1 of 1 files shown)


Xefyr0:
Update: The situation may be more confusing than I initially made it out to be.
It's not clear what the exact causes of the missing material info are.

In my experiments, GT Shaped Crafting & recipes that have any ingredients that *don't* have an attached MaterialInfo are less forgiving, so they should be higher priority and lower priority, respectively.

Additionally, it seems that tag ingredients (such as ALL the ingredients in the Thermal Machine Frame) do not work at all.

Marking this as draft - this may require creating some issues on GTM and further investigation.

Xefyr0:
Another update: 
There is indeed an issue with `addMaterialInfo` [that affects shaped recipes defined in KubeJS.](https://github.com/GregTechCEu/GregTech-Modern/issues/3641)

Most, but not all of the issues I've observed appear to be the result of this bug.

PR #951, opened 2024-10-16
Add optional compat for Super Factory Manager

juh9870:
This PR adds optional mod compat for the [Super Factory Manager](https://legacy.curseforge.com/minecraft/mc-mods/super-factory-manager) (SFM, for short).

# Balance
Recipes in this PR are gated around early HV (circuits, chemistry) but mostly with MV components and materials.

One major consideration is that SFM can transfer FE at an unlimited rate, and so to avoid gating it to the extreme end of the pack, this PR chooses to ignore FE transfer capabilities and balance it purely around item and fluid transport, but the actual merge of this PR should probably wait until SFM implements functionality to disable FE transfers: https://github.com/TeamDman/SuperFactoryManager/issues/174

# Recipe changes
<details>
  <summary>Click to expand</summary>

image

image

image

image

image

</details>

# Other changes

## Water Tank
Recipe is removed, pack already has enough means of water production

## Printing Press
Enchanted Book copying disabled, we already have EiO enchanter. Other uses (book copying, disk copying), as well as the press recipe itself are left intact

# Credits
Thanks `@.boold` on Discord for the initial script, and recipe ideas.

PR #2372, opened 2025-12-13
Add optional compat for Super Factory Manager (second try)

juh9870:
This PR adds optional mod compat for the [Super Factory Manager](https://legacy.curseforge.com/minecraft/mc-mods/super-factory-manager) (SFM, for short).

> This is an updated version of #951. SFM finally got the functionality to disable the transfer of certain resource types, so FE transfer is now disabled to not affect energy transfer balancing.
> 
> Recipes were not changed since the last PR. Client-side script was added to hide removed items.

# Balance
Recipes in this PR are gated around early HV (circuits, chemistry) but mostly with MV components and materials.

~~One major consideration is that SFM can transfer FE at an unlimited rate, and so to avoid gating it to the extreme end of the pack, this PR chooses to ignore FE transfer capabilities and balance it purely around item and fluid transport, but the actual merge of this PR should probably wait until SFM implements functionality to disable FE transfers: https://github.com/TeamDman/SuperFactoryManager/issues/174~~

FE transfer via SFM is disabled in server config

# Recipe changes
<details>
  <summary>Click to expand</summary>

image

image

image

image

image

</details>

# Other changes

## Water Tank
Recipe is removed, pack already has enough means of water production

## Printing Press
Enchanted Book copying disabled, we already have EiO enchanter. Other uses (book copying, disk copying), as well as the press recipe itself are left intact

# Credits
Thanks `@.boold` on Discord for the initial script, and recipe ideas.

### DIFF
--- a/kubejs/server_scripts/mods/optionalCompats/sfm.js
+++ b/kubejs/server_scripts/mods/optionalCompats/sfm.js
@@ -1,84 +1,84 @@
 // SFM compact script. Gated behind HV circuits and some chemistry, but recipes can be made in MV machines
 if (Platform.isLoaded('sfm')) {
-  console.log('SuperFactoryManager found, loading compat scripts...');
-  
-  ServerEvents.recipes((event) => {
-    // Remove most SFM recipes
-    for (const recipe of [
-      'sfm:manager',
-      'sfm:cable',
-      'sfm:fancy_cable',
-      'sfm:water_tank',
-      'sfm:disk',
-      'sfm:labelgun',
-      'sfm:network_tool',
-      'sfm:xp_goop',
-    ]) {
-      event.remove({ output: recipe });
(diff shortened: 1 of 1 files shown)


WithersChat:
Super Factory Manager is effectively Creative-Tier logistics. It erases any and all creative solutions to logistical problems that a player might encounter, and replaces them all with at most a simple programming assignment.  

The 0.13 update, and the pack's design direction from here on, are all about encouraging emergent gameplay and interesting logistical challenges. SFM as a mod goes against that ideal.
For that reason, **Monifactory will never support SFM**. Especially not as a MV unlock.

I would like to add that the main use case of SFM, being item routing to walls of HNN simulation chambers, has been addressed by the addition of overclockable HNN multis in EV.

PR #2373, opened 2025-12-13
revert oilsands removal from microminer missions

cewlboi:


WithersChat:
Ignore meI have blind.

PR #2377, opened 2025-12-14
Add the Monium easter egg

RubenVerg:
Monium, later renamed to [Victorium](https://en.wikipedia.org/wiki/Victorium), is a chemical that upon discovery was identified as a new chemical element; later it turned out to be a mixture of gadolinium and terbium. This PR adds a reference to this coincidence in naming, allowing monium dust to be electrolyzed into gadolinium and terbium. It also adds gadolinium dust, which (intentionally) has no intended use.


WithersChat:
Eh... 

Neat trivia but I don't want to add clutter to recipe lookups honestly.

PR #2384, opened 2025-12-18
remove unused files

cewlboi:
eio energy conduits are gone along with the mod that added them
T8.5MM assets
dense ores & related recipes
starry diamond block


kubejs/startup_scripts/Custom_Conduits.js
Xefyr0, line 1: Actually, the continued existence of this file means that custom conduits are still accessible for those who had them in their save and deleting it will remove those conduits from players' worlds, which is a breaking change that should be saved for a major version.

kubejs/startup_scripts/registry/block_registry.js
Xefyr0, line 1: If you are removing Dense ores, then remove their textures & model as well.

Xefyr0 (changes requested):

kubejs/startup_scripts/Custom_Conduits.js
cewlboi: we can just leave this for 0.14 or 1.0 then

kubejs/startup_scripts/registry/block_registry.js
cewlboi: they don't have models, but textures are gone now

WithersChat (approved):
Good to merge into dev. Xef?

kubejs/startup_scripts/registry/block_registry.js
```diff
@@ -135,7 +135,6 @@ StartupEvents.registry("block", event => {
         .tagBlock("forge:mineable/wrench")
         .requiresTool(true);
 
-
     // Dense Ores
```
Xefyr0: Looks like the Dense ore registry was left in by accident? If you're deleting the textures you should also get rid of their definitions.

kubejs/startup_scripts/Custom_Conduits.js
Xefyr0, line 1: Since you're removing the custom conduits, consider also removing KubeJS-EnderIO, the mod that allows for adding the conduits in the first place (and nothing else, AFAICT)

Xefyr0 (changes requested):

Xefyr0 (approved):

PR #2394, opened 2025-12-20
Allow blocks outside the Naquadah Refinery to be Predicates.any()

mosesyu1028:
Currently, there are two blocks (marked yellow in the first image below) that are set to be `Predicates.air()` when they are clearly outside the naq refinery. This causes the multiblock to break when a non-air block is placed there. This PR changes those blocks to have the rule `Predicates.any()`.

Those could be common spots for cables (GT power or AE2 or otherwise) which would break the multiblock for some users.
([Unless I'm the only idiot who couldn't figure it out](https://discord.com/channels/914926812948234260/1229854271613436066/1451773795462287380))

image

image


mosesyu1028:
PR is ready to merge :)

PR #2399, opened 2025-12-24
Tweak Compact Machines compat script to use GregTech materials and properly tier from HV

arsdragonfly:


Xefyr0:
Sorry, these alterations to the optional compat script are not what we're looking for.

Compact Machines have no need to use GT hull materials since... they aren't GT machines.

As for the tiering, it's intentional that the lower tiers are all unlocked at once and use lower tiers of components - 3x3x3 isn't much space especially before tunnels are unlocked, so it was made cheap.

PR #2403, opened 2025-12-26
Remove Gunpowder dupe fix

Xefyr0:
Undoes the patch we did for the dupe now that it's not necessary in GTM 8.0 thanks to [GTM#4301](https://github.com/GregTechCEu/GregTech-Modern/pull/4301)


Xefyr0:
Whoops. This should be draft.
Depends on #2423 being merged.

PR #2414, opened 2026-01-01
Update forge version to 47.4.10

NegaNote:
Adss java 25 support for better performance


WithersChat:
Is this properly tested?

NegaNote:
> Is this properly tested?

Tested to work with both java 17 and java 25.

Issue #2430, opened 2026-01-13
[Bug]: Phenol recipe for logs without nitrogen in the pyrolyse oven does not work


erencank:
### Issue Summary

I tried to make phenol in the pyrolyse oven with some rubber logs and circuit 11 following this recipe:

<img width="324" height="186" alt="Image" src="https://github.com/user-attachments/assets/7af29b2a-f4f1-481f-bc9f-e0284e612a96" />

However, it seems that this recipe is not working? There is a same recipe that uses circuit 11 which does require nitrogen, and if I supply nitrogen it does work.

<img width="316" height="185" alt="Image" src="https://github.com/user-attachments/assets/277c8e8a-a7e0-4687-b1eb-521f7227052e" />

 I am wondering if this recipe without nitrogen is supposed to exist or if the circuit might need to change?

### Reproduction steps

1. Make a pyrolyse oven with all the hatches and busses with lv power.
2. Put logs in with circuit 11
3. Nothing happens (Should start the recipe then?)
4. Provide nitrogen
5. Recipe starts

PR #2431, opened 2026-01-13
Change circuit value from 11 to 12 in pyrolyse oven for phenol logs

erencank:
Based on issue #2430, I went and made a MR to change the value of the circuit if the recipe is intended to exist. If this is not the case please feel free to close this MR!

I saw no other circuit here use 12 so I just chose that as a +1 over the old recipe. Not sure if there are any guidelines on what number to give to a circuit in a recipe.

closes: #2430


erencank:
> The circuit is changed on the wrong recipe. The nitrogen recipes always have a higher circuit number than any counterparts without a gas, so the log_to_phenol_nitrogen recipe should have been changed instead.

Ah I didn't pay attention to that, should be changed now.

Xefyr0 (approved):

PR #2453, opened 2026-01-26
Better Diagrams for Prismatic Crucible

MEEPofFaith:
- Created new diagrams which use images instead of text to describe arrows.

image

- Separated supercrit to it's own quest. Active and supercrit now have 2 separate diagrams.

image

image

image

- Added images to go alongside the examples in PSoC's description.
image

image

Yes, I was listening to Caramelldansen while making this.


WithersChat:
I wish I could merge only half of that PR. I love what you did with PSoCs but I'm not convinced by the rest.

WithersChat:
Closing this PR because I don't think I'll ever get decided on those diagrams and the whole cores lines need to be reworked in .14 or (most likely) .15

Thanks again for the PSoC diagrams, those were peak.

PR #2458, opened 2026-02-01
Update End_Game.js

Xefyr0:
Fixes a typo for EBFing Darconite wires for HDD wires where it consumes 1 wire instead of 4.


Xefyr0:
Duplicate of #2452

PR #2461, opened 2026-02-01
Allow Large Naquadah Reactor to only have laser hatches.

tomprince:
Previously, it required an energy hatch as well.


Xefyr0 (changes requested):
This PR allows the multi to be formed with up to 2 output energy hatches & 2 laser hatches, for a total of 4 output hatches. Instead, it should allow for 2 hatches of either type.

WithersChat:
Sorry for close, misclick.

WithersChat (changes requested):
Please make it so that there will only be a cap of two energy output hatches regardless of type.

tomprince:
No. That makes the discoverability of using laser hatches worse than the status quo, so I'd rather leave things in changed than that.

WithersChat:
Closing this. I'll have to do it myself.

PR #2476, opened 2026-02-08
Mod & Config updates

Xefyr0:
Mods excluded and the reasoning behind them:
GregTech Modern 7.5.0: Introduces an issue that can cause servers to crash.
WaterFrames and WaterMedia: Large backend change that requires a level of scrutiny not suitable for a clerical update like this PR.
Sophisticated Storage: Recipes changed to use Levers instead of Redstone Torches. Decided to postpone dealing with this change to a later date.


WithersChat:
Would be nice to list which mods you're updating for ease of review in the future, but it looks like you already made sure not to have any actual changes in there so we good.

PR #2489, opened 2026-02-15
Infinity Rebalance

Xefyr0:
What this PR does:
- Triples the output of all recipes that yield Infinity, making it twice as fast to get the same amount. Transmatrix input is also reduced to match Neutronium and Activated Netherite input.
- Adds 50% Transmatrix cost and 50% Null output to every Prismatic Core recipe
- Halve HoaU output of advanced HoaU mission to counterbalance increased Infinity and Null yields above
- Changes UEV Machine Casings to use one material like all Machine Casings before them.

How does this change things?
- PrismaC is much less of a bottleneck, because you don't need to produce as many Active Cores to satisfy your need for Infinity or Null
- UEV casings cost less despite the increased use of Infinity.
- Naquadah Superfuel and the Infinity Annihilation recipe for Antimatter now suck less, but are still subpar because the real issue with those fuels is the complexity and fragility of Infinity automation rather than the amounts.


WithersChat (changes requested):
This is a bandaid, but we need a bandaid for 0.14 because 0.15 is too far ahead.

The hyperbolic heart mission should also take only one active core instead of 2.
Also, if you feel like it, you could make liquid monium in the 24-ingot recipe go from 8 to 16mb. That's what I'd personally do, but I'll leave the decision up to you.

Once that's resolved, feel free to merge without asking me again.

PR #2493, opened 2026-02-16
Move PECA earlier and PEEK later in progression

Xefyr0:
Currently, PECA and PEEK are arranged oddly in progression. They come one after another, and despite being the more complicated line PEEK comes first. They're also part of a large pile of processing lines that can overwhelm the player across UHV.

This PR moves PECA earlier in progression as a requirement for UHV Machine Hulls and circuits, and moves PEEK later as a requirement for creative-tier multiblocks and MAX machines. This way, the wall of processing lines would not be so tall.

As a consequence of this rearrangement, several recipes and quests are changed including the cable coating recipes. It would be wise to review this PR thoroughly.


Xefyr0:
whoops, was on the wrong branch

PR #2494, opened 2026-02-17
PECA tweaks

Xefyr0:
Some tweaks to the PECA line to hopefully make it more enjoyable. Should synergize well with #2493.


WithersChat:
Woooaaahhh that one will need more eyes than me to review.

Xefyr0:
Don't be intimidated. This PR only does 4 things:
1. Remove Ethyl Cyanoacetate and Uncracked Ethyl Cyanoacrylate and associated lang
2. Merge the processing line steps that would otherwise have those two fluids as an intermediate
3. Increase the consumption of Dimethyl Sulfoxide so it's a non-negligible cost.
4. Tweak PECA polymerization to use less water and add an option for using Air instead.

WithersChat:
Nevermind. My eyes are enough. I'm getting better at this holy void.

PR #2495, opened 2026-02-17
Move some ore veins around

Xefyr0:
Moves the Wolframite and Briartite veins to Mars when they would be more helpful (Tungsten and Germanium respectively for IV prep)
Also moves the Coltan vein over to Mercury so the planets don't get too unbalanced - the Niobium from that vein is still available in time for the Niobium Titanium cost in Wetware.
Lastly, this PR moves the Tier 2 Rocket quest right after the first Signalum quest in Scaling up (rather than the second Signalum quest in Some Assembly Required)


WithersChat:
I'm merging either way but does that technically allow to skip CFA line for Germanium?

Xefyr0:
Yes, and that's intentional for several reasons:
1. Coal Fly Ash is bad. Not saying this as an opinion, it's a categorically bad line in its current form for several reasons:
1a. Low output of the desired resource (Germanium) for how much inputs are consumed (Coal) even after the buff.
1b. Byproducts (e.g. Goethite, Lithium, Aluminium) are undesirable due to being low-value or are more easily obtained through alternate methods.
1c. Coal Fly Ash is expected to be performed in EV before IV is made available, but is even larger than Normal Mode Platline which, as we know, caused adverse player behavior due to being a jump in processing line complexity.
2. Germanium is a low-value, low-consumption resource, which makes manual mining more viable compared to most other resources (including platinum group metals, to continue with the platline comparison)
3. The Germanium-bearing vein reduces to around a stack of Germanium on average, or 3 if you use the advanced miners. This would be sufficient for earlier tiers (EV/IV) but as you continue into the lategame, your need for Advanced Solder will continue to grow (especially once Optical circuits hit) and not all Advanced Solder recipes can be replaced with Living Solder. This means that players will be pushed to eventually switch to Coal Fly Ash line if they don't want to be interrupted by running out of Germanium every so often
4. Going back to Platline again, Germanium lacks the intimidating reputation Platline has, which should make players less averse to building Coal Fly Ash line when the need arises.

I do think a Silicon + Argon fusion recipe for Germanium would make sense, though.

PR #2500, opened 2026-02-20
Tweaks to combustion fuels' effectiveness

Xefyr0:
Boosts off-meta combustion options: Diesel, Biodiesel, CBD, and Rocket Fuel.
Also nerfs Naphtha slightly, but other fuels intended to be intermediate products are unaffected.

### DIFF
--- a/kubejs/server_scripts/fixes_tweaks/combustion_rebalancing.js
+++ b/kubejs/server_scripts/fixes_tweaks/combustion_rebalancing.js
@@ -1,0 +1,33 @@
+/**
+ * Replaces some base GregTech combustion recipes with more balanced versions.
+ * Well, "balanced" in a pack where synthetic Octane and JEAN exist.
+ */
+ServerEvents.recipes(event => {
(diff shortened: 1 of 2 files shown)


WithersChat:
Rocket fuel being a bad fuel is okay, it's fucking rocket fuel not turbine fuel.

As for Naphtha, why not just... remove its burn value outright?

Xefyr0:
Don't worry about Rocket fuel, it's still strictly worse than Diesel.

Rocket fuel generating 15A MV:
rocket_fuel

Diesel generating 15A MV:
buffed_diesel

As for Naphtha, the answer is Public Relations. The _apparent_ change of a buff or nerf is just as, if not more important than the _actual_ change. Completely removing Naphtha as a combustion fuel is bound to raise some pitchforks, but a nerf would be far more acceptable even though a 20% nerf to revenue translates to a 30-40% nerf in profit due to Naphtha's slim margins.

WithersChat:
I still think that killing Naphtha altogether and maybe buffing diesel a bit further for compensation is the right call.

WithersChat:
I still think that but we can talk about it later.

PR #2501, opened 2026-02-20
Nuke Americium Plasma.

Xefyr0:
Putting this here so I don't forget about it later.


WithersChat:
Sad to see it go instead of improved but probably for the best.

PR #2510, opened 2026-03-04
Remove the need for NBTs on Thermal Augments

Blackalf:
Modify the Thermal Augments Item Properties on startup instead of adding NBTs to recipes.

Players can now successfully pick up augments from EMI
All previously crafted augments will be automatically updated if their stats are changed (Unless they already have NBT)


Blackalf:
I tested it on my world and could properly get the augments with the right stat buffs using EMI Cheat Mode or by crafting the items. They also properly updated when modifying the augment's stats.

Temporary 1/1 recipes to reset old augment's nbts could be added if wanted

WithersChat:
Please add those recipes 🥺

Blackalf:
All done! :D

Xefyr0 (changes requested):
Looks great.
Since this gets rid of the only use of `thermal.emi.tooltip` would you mind getting rid of that too in each lang file? They're located in `kubejs/assets/kubejs/lang/`.

PR #2516, opened 2026-03-07
Chemical bathing recipes for (mixable) dyes

iwishforducks:
Adds Chemical Bathing recipes for vanilla minecraft dyes from GregTech's Chemical Dyes with a Slime Ball.

This is to provide an alternative for creating and mixing dyes rather than cycling flowers through the Atomic Reconstructor (and is not meant to be a replacement for such)

I decided on adding the recipe to turn the dyes back into the minecraft counterparts rather than adding a mixing recipe between the chemical dyes since "chemical dyes" aren't usually used for mixing colors together.


Xefyr0:
Why did you make this PR? Under most circumstances, the GT Chemical dye ball items, made by solidifying the liquid form, can be substituted for vanilla dyes. Moreover, those chemical dyes can be produce indefinitely without growing any plants by cycling the solid and fluid forms while providing sulfuric acid and salt.

PR #2517, opened 2026-03-07
Implement Wither Gas

Xefyr0:
Creates a Wither-themed Gas Turbine fuel counterpart to JEAN Gasoline.
Includes a short chemline (4-5 steps in most cases assuming Nitrobenzene is automated) to produce the fuel.


WithersChat:
Do we want this as part of a patch or the powergen branch?

NegaNote:
seems better for powergen branch imo

PR #2521, opened 2026-03-12
change wire used in iv singleblock electrolyzer

cewlboi:
the rest of the iv singleblocks that use wires use rtm alloy wires


Xefyr0:
RTM alloy is reserved for heating elements. Changing Electrolyzer wire material should be done in `gregtech_crafting_components.js`.

PR #2523, opened 2026-03-13
Remove Infinity from Naquadah Superfuel

Xefyr0:
Self-explanatory.
There's probably something to say about the Infinity antimatter recipe, antimatter can be used to make power without it so it's lower priority.


WithersChat:
Not 100% sure this is all we need but it's a good start.

PR #2524, opened 2026-03-14
Add Purpur processing

Xefyr0:
Adds processing for Purpur, which provides passive production for Potassium, Manganese, Phosphorus, and Ender Pearls.

This should be thought of as something between Ender Spores and Dragon Scale processing.


WithersChat:
Looks good. Makes shulker shell processing potentially more worth it as well, which is always appreciated.

PR #2527, opened 2026-03-15
Optical Matrix

Xefyr0:
Changes the Crystal Matrix processing line to take what's unique from Optical Chips, then gets rid of Optical Chips so that Optical Circuits use Photonic SoCs directly when they're made for the first time.


WithersChat (changes requested):
Might want to make Optical Processors craft in batches of 4 to not throw a wrench in circuit progression costs.

Especially since this change suddenly doubles the PSoC demand, and therefore the PrismaC stress.

Xefyr0:
> Might want to make Optical Processors craft in batches of 4 to not throw a wrench in circuit progression costs.
> 
> Especially since this change suddenly doubles the PSoC demand, and therefore the PrismaC stress.

Done.

PR #2534, opened 2026-03-22
Move some HNN loot to more appropriate mobs

Xefyr0:
List of changes in this PR:
- Copper ingots moved from Spider to new Drowned model
- Gold ingots moved from Guardian to new Zombified Piglin model
- Spider model yields Zinc to replace Copper
- Nickel yield from Slime model buffed
- Gunpowder and Coal yields from Creeper model reduced, with Ghast and Wither Skeleton offering alternate sources to those resources respectively


WithersChat:
I like playing into modern mobs but... isn't that almost bloaty? 

Not to mention the guardian model being even more useless now.

Xefyr0:
It might be close, but it really bothers me that HNN doesn't reflect the mobs that actually yield Gold and Copper as drops.

Fortunately, the difference in a practical setup is only a matter of two machines since Loot Fabricators are the bottleneck, not Simulation Chambers. Once you move beyond Simulation chambers the difference is even smaller because you're reallocating some simulation computers used for other models to the new models rather than making new simulation computers due to hatch limits.

As for the Guardian model, it's a little concerning that you consider it to be "useless" despite the fact that it has the most options for pristine matter fabrication. I contend that Wet Sponges at least are useful for clearing modern worldgen's underwater caves (even though their frequency has been reduced) Cod is also useful as a simple source of Mince Meat for NM Wetware (which dev run used a simple TE Fisher setup for)
I do agree that Aluminium and Guardian scales could be pretty useless thanks to Clay electrolysis and poor balancing respectively though. I'll try rebalancing Guardian Scales in a separate PR, LMK if you come up with ideas for making Prismarine useful for something more than building or think Aluminium should be yote from the loot pool.

Xefyr0:
~~On a related note, I had a thought - in addition to rebalancing Guardian Scaleline we could merge it with Prismarine (i.e. Guardian Scale Line gets rebalanced and has Prismarine as an input instead of Guardian Scales)~~

EDIT: Done in another PR

WithersChat:
I do think that yeeting aluminum from the model would be good. It's a pure noob trap as of now. 

As for the Prismarine issue, see my comments on that PR.

PR #2537, opened 2026-03-25
Rework Guardian Scaleline into Prismarine Processing

Xefyr0:
Opted to put this on Main instead of Processing Lines Reshuffle since Guardian Scaleline is bad enough that this could be considered a "fix"


WithersChat:
Quick question: What did you change in the line inputs and outputs? (Beyond the whole scale->prismarine change)

WithersChat (changes requested):
Prismarine Crystals and Shards are still obtainable from the Atomic Reconstructor.

Please remove both those recpies, and instead have one AR recipe turning shards into crystals for builders.

PR #2548, opened 2026-04-02
Properly adds Phoenix's Chromatic Codes

Phoenixvine32908:
I'm uhhh, really sorry about last time lol.

### DIFF
--- a/build/index.ts
+++ b/build/index.ts
@@ -46,7 +46,8 @@ const clientMods = [
     "drippyloadingscreen",
     "badoptimizations",
     "fancymenu",
-    "gtmoldraw"
+    "gtmoldraw",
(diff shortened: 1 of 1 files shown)


Xefyr0:
If I'm understanding this mod properly, it will serve no purpose until used in the questbook or other text.

I see no point in adding this mod without using it. It will be bloat otherwise.

PhoenixOnTheVine:
I plan to use it for the emi integration for the teal of prismaC to make it actually match prismaC and withers said it would be ok to give the svat a green that belongs to it.

Xefyr0:
OK, I'll be happy to add the mod when that is ready.

Phoenixvine32908:
Both added, oh and the colors
image
image

PR #2551, opened 2026-04-04
Rebalance Turbine Rotors

Xefyr0:
Buffs turbine rotor durability across the board.

Objective:
Durability values should result in turbine rotor replacement playing like the maintenance mechanic: By the time your rotors start to break, the average player would have unlocked a better material to make rotors with. Exceptions should be limited to players that progress particularly slow, players that are on servers, players that poorly manage their turbines' duty cycles, and players that deliberately use rotors with lower durability values for higher efficiency or power output stats. Even then, replacing broken rotors with new ones of the same material should only happen a few times per playthrough rather than a few times per voltage tier.

The guiding principles I used to balance this are as follows:
1. Buff rotors with low base durability, low efficiency, and _especially_ rotors with low power output harder than rotors with high stats.
2. All rotors should get a net increase in durability, except Neutronium

Desmos graph used for calculations:
https://www.desmos.com/calculator/itkapmavv6

### DIFF
--- a/kubejs/startup_scripts/gregtech_material_registry/material_modifications.js
+++ b/kubejs/startup_scripts/gregtech_material_registry/material_modifications.js
@@ -135,6 +135,7 @@ GTCEuStartupEvents.materialModification(event => {
     // We keep Ingots in the material definition so we can replace it in the Ore Processing Diagram with vanilla Netherite Scrap, then remove it here.
     TagPrefix.ingot.setIgnored(GTMaterials.get("netherite_scrap"), Ingredient.of("minecraft:netherite_scrap"))
 
+    // Reduce the abnormal spike in base GTM pipe properties as progression continues
     GTMaterials.Neutronium.getProperty(PropertyKey.FLUID_PIPE).setThroughput(400)
     GTMaterials.Neutronium.getProperty(PropertyKey.FLUID_PIPE).setMaxFluidTemperature(10000)
     GTMaterials.Ultimet.getProperty(PropertyKey.ITEM_PIPE).setTransferRate(4)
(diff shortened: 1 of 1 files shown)


WithersChat:
Nerfing Neut durability when it should have been unbreakable from the start is a bit :skull: NGL

And overall, why do we even want rotor durability? Maybe erasing it entirely would make steam and gas slightly more attractive to players.

Xefyr0:
Because turbine durability (and the material costs associated with it) should be a key balancing element for turbine rotors. Remove it and you remove all depth to the mechanic - at that point so many rotors become noob traps that you may as well remove turbine rotors in their entirety.

The sad part is that due to the ripple effects of previous balance decisions, we're nearly there already.

The primary factor for most players when considering rotor selection is the **power output**. The other stats don't come even close, and durability comes last in terms of importance. From the player perspective, efficiency is a non-factor since more power enables you to scale more fuel, and durability is ignored because investing in it yields small returns. For all rotors unlocked after platline but before naqline, the most durable rotor only nets you 3x as much durability as the least durable rotor, while the rotor with the most power output yields **2x as much power** as the least powerful rotor. When resources come rolling in without the player having to lift a finger, the increased demand on any turbine autocrafting system is unimportant compared to building twice as many turbines when turbine-based generators are already suffering from a need for spam.

And yes, the decreased importance of efficiency and rotor durability from a player's perspective compared to what they should be is directly correlated with Moni's penchant for handing out copious quantities of renewable resources. Go figure.

These skewed player priorities results in a very narrow meta. Currently there are maybe ~4 rotors in that meta, and removing durability only entrenches them further since durability would go from negligible down to a literal non-factor.
That "meta" progression is the following:
Ultimet -> Tungsten Carbide (Optional) -> HSS-G (Optional) -> HSS-E or HSS-S -> Neutronium.
I even added in the optional rotors since they're technically better than Ultimet but unlock later - %t turbine_rotor recommends players to use only 3 rotors out of the 35 available and your idea of removing durability would **not help**.

Instead, rebalancing rotors to make investing in durability worthwhile would open up that meta. With higher-durability rotors offering so much run-time that they last until the next upgrade, investing in durability in leiu of power still means players have to build additional turbines, **but it also means that they don't have to build an automatic rotor replacement system.**

Now, after reading this you may be inclined to suggest that we make rotors with low power output unbreakable instead. After all, if additional durability is good for rotors with low power output, why not take that to the logical extreme?
The answer is simple: all good things in moderation. Making some rotors unbreakable but not all will swing the pendulum too far to the other side - players will go back to optimizing for power, but this time they'll only consider the unbreakable rotors instead of the full 35, just like how players only tend optimize for cost yet only consider superconductors when it comes to wiring.

Xefyr0:
In retrospect, the essay in my previous comment may appear to conflict with the fact that this PR doesn't solve the balance issues among turbine rotors either.
As explained in the first comment, all this PR does is reduce the impact of turbine rotors' limited durability. (the primary drawback I've seen players list when comparing Turbines to Combustion)
Eliminating durability as a mechanic would preclude the possibility of ever balancing the wide selection of turbine rotors available, so while this PR doesn't necessarily solve the problem it leaves the door open.

WithersChat:
Closing this since we removed rotor durability since. We should probably start the balancing process from scratch.

PR #2559, opened 2026-04-09
Make Germanium easier

Xefyr0:
This PR makes the player experience surrounding Germanium easier by adding a fusion recipe for Germanium and drastically reducing Germanium consumption for alternate Diode recipes.

The one area this PR does not change, unfortunately, is the area that it's needed the most: Coal Fly Ash line.
CFA is longer than Platline, required at nearly the same time, and its products are less useful outside of Moni-specific progression gates. Coal Fly Ash is overall a poor choice for Germanium production in its current state & location in progression. There is currently an ore available for Germanium in EV which is helpful as a workaround (and a popular choice for that reason) but I am in need of ideas on how to improve the situation.

As a side note, I can understand that fixing CFA line may make this PR redundant - if action is taken to fix CFA, I don't mind having this PR closed as it's mostly a band-aid to cover up a deeper issue.


WithersChat:
We need to fix CFA in a 0.14 branch and to remove the fusion recipe then, but this should help for now.

PR #2573, opened 2026-04-13
Add fusion recipe for Iridium

Xefyr0:
Recipe uses Tantalum and Beryllium.

If the recipe is left unused, switch to a cheaper option from the original poll: https://discord.com/channels/914926812948234260/1229929078547550238/1492714110943494196
If this results in the Ir/Os ores on Glacio becoming useless, consider buffing them to be layered veins instead of veiny veins.

If neither of the above occurs, good.


WithersChat:
If it stays unused it will be because people skip on other parts of infra.

Beryllium is something you should be getting a lot of as a side from processing other stuff, and tantalum is 4 machines to recycle something you're getting a lot of from breath and eggs with MMs. It's not free, but between scaling this and scaling platline, scaling this is easier.

And the people who love skipping on infra will mine as they always do.

PR #2576, opened 2026-04-18
Force upright Greenhouses

Xefyr0:
Currently, Greenhouses have a render that grow crops atop the dirt blocks, but can be rotated any which way. This results in the renders sometimes creating unexpected or unrealistic visuals.
image
This PR forces Greenhouses to form upright like Distillation towers to ensure the renders are always visible on the inside of the greenhouse and grow upwards from the dirt, just like real greenhouses.


Xefyr0:
If forcing a particular multiblock orientation is not acceptable, then the Greenhouse render should rotate with the multiblock orientation instead.

WithersChat:
People won't like this, but I do /lh

PR #2593, opened 2026-05-07
Tweaks to AE2 recipes

Xefyr0:
This PR does 3 things:
1. Tweak Requester and Matter Condenser recipes, adding a Fluix plate to the former and making the latter more interesting.
2. Change gregified AE2 Processor recipes definitions to better adhere to the DRY principle
3. Change most 'magic numbers' for power in the AE2 recipe file to use energy constants from GTValues.

### DIFF
--- a/kubejs/server_scripts/mods/AE2.js
+++ b/kubejs/server_scripts/mods/AE2.js
@@ -1,5 +1,5 @@
 /**
- * Custom recipes for AE2
+ * Custom recipes for AE2 & addons
  */
 ServerEvents.recipes(event => {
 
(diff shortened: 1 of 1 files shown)


WithersChat:
Can you explain what point 2 did? And also why do those still use magic numbers?

Xefyr0:
The effects of point 2 are in lines 850-875 for the file - it's just a switch to using one recipe definition with fields that are populated using info from a data structure rather than several similar definitions one after another.

As for "Magic numbers" it's just that some recipe definitions used `.EUt(30)` instead of `.EUt(GTValues.VA[GTValues.LV])`. The two are equivalent, but the latter is more transparent about what the function call is intended to do - it's like talking about energy in volts & amps instead of raw EU/t.

WithersChat:
I know what magic numbers are, you just forgot yo replace a few in the code for part 2.

Xefyr0:
Right. I didn't do them at first because the values didn't directly correspond to any numbers in GTValues, but that's what coefficients are for :V

PR #2594, opened 2026-05-07
Nuke NCN moderators

Xefyr0:
I'm aware that NCN is set to be gotten rid of in favor of Phoenix Fission, but I've found evidence to suggest that some of NCN's spontaneous reactor explosions are the result of moderator heat calculations changing for some unknown reason.

Thus, for those who don't intend to migrate to Phoenix Fission for any reason, this PR removes moderators so players don't stumble into that issue.

### DIFF
--- a/kubejs/server_scripts/mods/NuclearCraft.js
+++ b/kubejs/server_scripts/mods/NuclearCraft.js
@@ -104,16 +104,6 @@ ServerEvents.recipes(event => {
             .addMaterialInfo(true)
     }
 
-    function canmod(name, input) {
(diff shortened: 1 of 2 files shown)


WithersChat:
...you **do** know that we're removing NCN as a whole in the same update we're adding Phoenix Fission, right?
This patch is only for people who won't touch 0.14.

Xefyr0:
Yes, that's the point.
People will stick with the last 0.13.x update for any number of reasons, and if they plan to use NCN then I'd like to carve out the buggiest parts to smooth their experience.

PR #2596, opened 2026-05-10
Update descriptions for Biodiesel and Glycerol quest

TheConfusdGoodra:
Added a line in the Glycerol quest for Ethylene production, highly recommended to switch to the glycerol route in HV since it's 2 buckets of ethylene and not colossally slow.


WithersChat:
It isn't, in fact, highly recommended. Glycerol is bad, and petrochem distillation gives much better rates.

PR #2602, opened 2026-05-14
Use AE2 config to hide facades instead of via KJS

Xefyr0:
Self-explanatory.

Biggest risk of this PR is that players' client configs will be overwritten each update, but we discourage updating in-place due to issues like this anyways.


WithersChat:
This will still reset configs of anyone changing modes which is... suboptimal.

WithersChat:
Can you remind me what the benefit of this even is?

Xefyr0:
I not for the fact that it would reset users' configs each update, it would be more elegant to do it this way (and potentially result in faster load times) but in retrospect I don't think the pros outweigh the cons anymore

PR #2603, opened 2026-05-14
Add HoaU to the PrismaC recipe

Xefyr0:
Currently the HoaU feels a little separate from the rest of the progression, only becoming a relevant gate at the end of UHV when the player aims to make Infinity/Null. This adds it to the PrismaC recipe so it has more weight in all modes, unlocking a multiblock that's going to be the core of progression (and the source of plenty of struggling :trollface:) for the remainder of the game.


Xefyr0:
Leaving this PR as draft because Withers was experimenting with moving the PrismaC earlier in progression which would conflict with this PR.

WithersChat:
PrismaC pre-heart won't happen until 0.15, so this is good until then.

PR #2610, opened 2026-05-28
Greenhouse fun

Xefyr0:
Lotsa stuff in this one.
- Greenhouse recipes now use varying amounts of water (aquatic plants use more, Cacti use less)
- Reduced output of growing cosmetic plants and flowers in the Greenhouse somewhat
- Added Glow Shrooms, a renewable source of Glowstone
- Phytogenic Insolator recipe for Ender Spores removed
- Ender Spores and Glow Shrooms made available in NM
- Added many more permutations of Fertilizer recipe
- Nerfed Mixer recipe for Fertilizer
- Reworked Phyto-Gro to be a more potent fertilizer than regular GT Fertilizer, with a more complicated recipe
- Made the recipe for the Phytogenic Insolator cheaper

In the end I decided against adding dimension requirements - it would only be meaningful for the Chorus flower, (other plants are not necessary or helpful for progression) and the only relevance the Chorus Flower has beyond building is for Warp Stones which are QoL.


NegaNote (changes requested):
Get rid of the added lang entries for german that were missed when resolving conflicts, and it's good

NegaNote (approved):

PR #2613, opened 2026-05-31
Make RTM coils less skippable

Xefyr0:
Does several things to make RTM coils more worthwhile to invest in on the way to HSS-G:
- Shorten RTM alloy's blast time 
- Replace Tungsten in HSS-G coils' recipe with Ruthenium
- Changes Tungsten and its alloys to require RTM alloy coils or higher
(RTM coils are needed for Tungstensteel in Nomi CE, after all)
- Make Molybdenum and Tungsten (Key elements of HSS-G) slower to get from Microminers


WithersChat (approved):
Approved, but will merge after the hotfix things are done.

PR #2614, opened 2026-06-02
PTFE Patterns 2: Electric boogaloo

Apoptosis847639:
Added an EV circuit + PTFE pattern recipe for quadruple the yields of patterns.

### DIFF
--- a/kubejs/server_scripts/mods/AE2.js
+++ b/kubejs/server_scripts/mods/AE2.js
@@ -87,14 +87,14 @@ ServerEvents.recipes(event => {
     }).id("kubejs:ae2/blank_pattern")
 
     // PTFE Patterns
-    event.shaped(Item.of("ae2:blank_pattern", 32), [
+    event.shaped(Item.of("ae2:blank_pattern", 16), [
         "AAA",
         "BCB",
         "BBB"
     ], {
         A: "gtceu:fine_silver_wire",
         B: "gtceu:polytetrafluoroethylene_plate",
-        C: "#gtceu:circuits/hv"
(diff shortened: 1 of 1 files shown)


Apoptosis847639:
also wrong branch name :sob:

WithersChat (approved):
Nice IM reference!

kubejs/server_scripts/mods/AE2.js
```diff
@@ -86,6 +86,17 @@ ServerEvents.recipes(event => {
...
 
+    // PTFE Patterns
+    event.shaped(Item.of("ae2:blank_pattern", 32), [
+        "AAA",
+        "BCB",
+        "BBB"
+    ], {
+        A: "gtceu:fine_silver_wire",
+        B: "gtceu:polytetrafluoroethylene_plate",
+        C: "#gtceu:circuits/hv"
```
Xefyr0: This ain't an EV circuit dawg 💀
Also, maybe do 16 or 20 patterns instead of 32? If you're going with an EV circuit.
For Crushed circuit in Transmatrix line, the yield from each successive circuit tier is double the last and yet it's still worth doing higher tier circuits so it should be the same here.

Xefyr0 (changes requested):

kubejs/server_scripts/mods/AE2.js
Apoptosis847639: I coulda sworn it was an EV circuit, another apop %t bold incident, I'll set the yield to 16

Xefyr0 (approved):

PR #2615, opened 2026-06-02
PPM Autoclave recipe 2: Electric Boogaloo

Apoptosis847639:
Reduced speed of autoclave PPM recipe to 1 second


WithersChat (approved):
Neat.

PR #2630, opened 2026-06-07
Updated Russian localization for quests

Hramoff:


Xefyr0 (approved):
I can't read Russian but translations for Moni can be made freely on Weblate so I'll assume that this was made in good faith.

PR #2634, opened 2026-06-11
Add casing blocks to multiblock machine quests

Xefyr0:
Puts casings and hatches in FTBQuests tasks for quests that don't already have them so players can get a rough idea of how much the full multi costs and the hatch configuration to use for the first build.
All casing tasks, both existing and new, are optional with the exception of the EBF and Pyrolyze Oven quests.

This PR also moves the PSS and Centralized Power quests to Scaling up in a bid to avoid merge conflicts with #2633.


WithersChat (changes requested):
Please make EBF and pyro oven casing task also optional because returning players sometimes start placing blocks before the rest is crafted.

Feel free to merge as soon as that's done.

WithersChat:
I appreciate your efforts to limit merge conflicts. It sadly didn't entirely work.

Still managed to fix it by hand in the end.
I think I'll do the optional task change myself to save you headaches.

WithersChat (approved):

PR #2636, opened 2026-06-12
LaserIO Card Update

random-account-name:
Increases LaserIO stack size to 16 and adds assembler recipes back to work with new stack size.


WithersChat:
That would be problematic with the way LIO UEL is coded.

PR #2638, opened 2026-06-18
Removed HNN req. from nether cake in NM

Apoptosis847639:
Self explanatory I think...


WithersChat:
Just make the HM recipes exist in all modes.

WithersChat (approved):

PR #2640, opened 2026-06-19
nuke ev+ converters and lumium+ LIO overclockers

PlasmaticVoide:
Adds all converters of tiers EV and above to nukelist
Adds all LaserIO overclockers of tiers Lumium and above to nukelist
Cryolobus overclocker used in Dark Soularium Thruster replaced with cryolobus wire

### DIFF
--- a/kubejs/startup_scripts/nukeLists/item.js
+++ b/kubejs/startup_scripts/nukeLists/item.js
@@ -138,6 +138,7 @@ global.itemNukeList = [
     "gtceu:luv_solar_panel",
     "gtceu:zpm_solar_panel",
     "gtceu:uv_solar_panel",
+    /^gtceu:(ev|iv|luv|zpm|uv|uhv|uev|max)_\w+a_energy_converter/,
     /^gtceu:uiv/,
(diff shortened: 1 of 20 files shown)


WithersChat (approved):
Nice.

kubejs/startup_scripts/nukeLists/item.js
```diff
@@ -211,6 +212,7 @@ global.itemNukeList = [
     // LaserIO
     "laserio:logic_chip",
     "laserio:logic_chip_raw",
+    /^laserio:energy_overclocker_card_tier_[5-9]/,
```
Xefyr0: This is the wrong way to get rid of the overclocker card items.
Instead, remove the config entries that generate the overclocker cards in `laserio-common.toml`.

Xefyr0 (changes requested):

kubejs/startup_scripts/nukeLists/item.js
Xefyr0: You'll also need to get rid of the lang entries for those overclocker cards.

PR #2661, opened 2026-07-03
Migration to Omicron Industries

WithersChat:
Okay, I'm gonna need help on that one, feel free to PR to this branch. Need to update all the links in the pack as well as any github actions or other stuff that might break.


WithersChat:
Okay so having an extra review would be neat but beyond that all we're missing is the discord link. Peak.

config/fancymenu/customization/moni-main.txt
Xefyr0, line 1: As mentioned in this comment thread, the Discord invite link needs to be changed.

Xefyr0 (changes requested):
There are still a few places where Pan is listed as the (singular) author, but they should be left alone in this PR for several reasons. There's also a link to his Patreon but it's less clear to me what should be done about that, if anything.

Aside from the Discord invite link, I think this is ready.

WithersChat:
I am quite sure that I deleted the patreon link ages ago? Did I fail to?

But yeah the other places we'll update probably with 0.14?

PR #2664, opened 2026-07-05
Nuke GT World Accelerators

Xefyr0:
With balance changes to endgame processing lines, these are no longer necessary and exist only as a vector for players to unintentionally lag out their game.


Xefyr0:
Need to do quests that cover WAs

Issue #2667, opened 2026-07-06
[Bug]: Quest book factually incorrect about cables


HarvT:
### Issue Summary

The new cable covering quest in scaling up says that both silicone rubber and SBR can cover all future cables "can wrap any cable tier" but silicone rubber cannot in fact wrap any UEV cable and can only do cryococcus at UHV, the only option in these cases being SBR. Im unsure if this is intentional or not so don't know whether it is an old quest line that needs updating or if come cables are missing recipes.

A similar thing is seen with the polyphenylene sulphide quest saying it is required for wrapping "LuV+" cables but again only works with cryococcus at UHV and nothing at all at UEV. 

This may occur elsewhere but I have checked obvious places like pvc wrapping which to my knowledge is not mentioned at all in the quest book despite being optional until EV machine hulls so a mention there could be helpful

### Reproduction steps

Both quests are mentioned above but for completeness
1. New cable coverings quest inaccurate
2. Polyphenylene sulphide quest inaccurate

### Relevant log output

```shell

```

PR #2670, opened 2026-07-10
Polyphenylene Sulfide Quest Fix

HarvT:
Polyphenylene sulphide is required to wrap cables from luv to uv inclusive but only works on cryococcus at uhv while all others at uhv and future tiers need PEEK.

This only fixes the en_us file as I don't know how you guys handle translations, feel free to copy or add the other translations accordingly.

Should fix the 2nd part of #2667 (the current pr only fixes the rubber coating inconsistencies not the plastic wrap).

Language is modified form xefyr's fix to the first part of #2667 for consistency.

Also the colours are fixed and LuV and UV are now their respective colours not dark green. 

Has been tested in game seems fine :)

### DIFF
--- a/kubejs/assets/ftbquests/lang/de_de.json
+++ b/kubejs/assets/ftbquests/lang/de_de.json
@@ -2144,7 +2144,7 @@
     "moni.quest.6B908DECFA46C8E0.description2": "It can also optionally overclock recipes up to 128 EU/t, gaining additional speed for relatively more power.",
     "moni.quest.6B908DECFA46C8E0.title": "&2HV Schaltkreis-Montageeinheit",
     "moni.quest.6BBACA979DF9BA30.title": "debug: undo hm",
-    "moni.quest.6BE1CD77618EC6C7.description1": "&2Thin Polyphenylene Sulfide Sheets are required for wrapping LuV+ cables.",
+    "moni.quest.6BE1CD77618EC6C7.description1": "&2Thin Polyphenylene Sulfide Sheets are required to wrap any Cable tier from &dLuV&2 up to &3UV&2.&r",
     "moni.quest.6BE1CD77618EC6C7.title": "&2Polyphenylene Sulfide",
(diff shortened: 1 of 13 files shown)


Xefyr0 (approved):

Xefyr0 (changes requested):
For quests, typically we update untranslated lang keys for other languages too (i.e. translation keys still in English) so that when players use other languages, they have up-to-date English defaults for things that haven't been translated yet. (as opposed to out-of-date information or completely blank text fields)

In this case, it's needed for every language except Russian and Chinese (Simplified)

HarvT:
Done :)

Also out of interest what's the best way to help out/learn pack dev for this sort of pack? I have a couple months free before I head back to uni and I love this pack so would be nice to help out even as a small contributor :)

Xefyr0 (approved):

Xefyr0:
If you want to help out, there's a lot of clerical work that needs to be done in the QB. It doesn't reach most of the quality standards described in [CONTRIBUTING.md](https://github.com/Omicron-Industries/Monifactory/blob/main/CONTRIBUTING.md) particularly in reference to highlight color since CONTRIBUTING.md was updated with new standards without changing any quests.
Accuracy of information is also an issue since there have been a lot of recent changes that affected gameplay but didn't change the QB to match.
This work would look very similar to what you've accomplished in this PR.

For learning pack dev, IMO the best way is to learn by doing. Pick a feature you or others want to see in the pack and see if it can be implemented using the techniques demonstrated in this repo. Keep in mind that not all ideas can be implemented without writing a mod in Java, and some common ideas have been previously suggested and rejected for balancing reasons or for going against the creative direction we want to go with this modpack.

HarvT:
Okay perfect thank you so much! 

Unfortunately I'm out of town for about a week but when I'm back I'm happy to do some grunt work with the qb while toying around and learning the internals a bit better. I even have had some minor ideas for a while that might be nice to implement locally and see if people want them upstream but that will be down the line a fair bit.

Might see if I can take my laptop and have a look the the qb to see if anything glaring sticks out. Although can't push until back as my laptop cannot run the pack and I'd rather push code I've tested in game.

PR #2674, opened 2026-07-13
Rotor rebalance 2

Xefyr0:
Turbine Rotor rebalance to go with the removal of the durability mechanic.

This time, I used an Excel spreadsheet to visualize how different rotors' stats compared against one another.
Picture1

This PR also removes and hides a collection of rotors that are rarely or never used using a technique new to this repository: EMI offers a way to hide itemstacks using a resourcepack. This technique is necessary since all turbine rotors are the same item but with different NBT.

There's no way to make the Activated Netherite turbine rotor have the enchantment glint other Activated Netherite items have unfortunately. LMK if that's an issue and I can remove the new rotor from the PR.

### DIFF
--- a/kubejs/assets/emi/index/stacks/rotor_rebalance.json
+++ b/kubejs/assets/emi/index/stacks/rotor_rebalance.json
@@ -1,0 +1,16 @@
+{
+  "removed": [
+    "item:gtceu:turbine_rotor{GT.PartStats:{Material:\"gtceu:magnalium\"}}",
+    "item:gtceu:turbine_rotor{GT.PartStats:{Material:\"gtceu:wrought_iron\"}}",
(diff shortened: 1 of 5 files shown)


Xefyr0:
While this PR should be a net positive for Turbine generators, it may be possible that they remain lackluster either due to how conservative this buff is or due to other, yet-to-be-released factors that counteract this PR (Namely, the nerf to Helium Plasma)

In such a case, it may be prudent to buff turbine generators across the board. Such a buff could be accomplished through rotor bonuses, but there are other options so there should be a discussion about the best implementation for such an objective.

WithersChat:
I am not worried about plasma, but I am worried about this being insufficient to buffing gas turbines (and maybe Steam too).

WithersChat:
Eh, you know what? We can rebalance that shit in .14.x

PR #2677, opened 2026-07-14
Remove the part about bleeding edge builds in FAQ

esmenard:
I guess you planned to have bleeding edge builds on the Github Releases page at some point, but only the stable releases (the same as in Curseforge) are there.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -71,7 +71,6 @@ By default, Monifactory only supports the mods included with it, and has *option
 **Q. How do I download Monifactory?**
 
 Stable releases of Monifactory can be downloaded from CurseForge using any launcher of your choice (PrismLauncher, CurseForge, etc.)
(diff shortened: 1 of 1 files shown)


Xefyr0:
This is true. Paging Withers on this matter - we *could* start doing bleeding edge releases (or plan to start doing them past some future milestone) or alternately merge the PR and remove the text.

Not sure which one would work better given our workflow and habits.

WithersChat:
Our current habits are better suited for having bleeding edge builds accessible via Discord, since we explicitly want players using BE builds to be able to give us direct feedback. This is what we did with Moni Meowdays and I think we should keep it that way.

WithersChat (approved):

WithersChat:
Requesting a Xef review to my changes, but this is my stance.

Xefyr0:
LGTM

PR #2679, opened 2026-07-16
lowered min casings on Large Engraving Laser 50 -> 44

WhitePhant0m:
lowered minimum casing count on Large Engraving Laser to allow exactly enough input busses for all 16 lenses.


WithersChat:
Ae2 mfs on their way to scientifically construct ways to never use more than 1 machine.

(This is not needed, especially on one of the slower machines in base GT.)

PR #2680, opened 2026-07-16
added pch to pyrolyse oven and cracker

WhitePhant0m:


hijiol:
Industrial pyrolysis Chamber my beloved

Xefyr0:
Please explain how this would be a net benefit.
Faster machines are not always better - you will need to demonstrate that Crackers and Pyrolyse Ovens are significantly slower than other machines or are in need of a buff. "They're slow" is not enough of a justification.

If it's only one application of these machines that is causing a problem, then it is easier to justify altering the recipe rather than the machine as a whole.

WhitePhant0m:
I made this PR as Withers requested it. I myself think that having to spam crackers/pyro ovens is annoying, which is why I'd like this, but main reason for this PR is Withers.

WithersChat:
Out of curiosity, how many pyro ovens/crackers did you have to place **outside of transmatrix line**?

WhitePhant0m:
I had around 32 pyro ovens doing charcoal byproducts/coal tar/wood tar/etc and 64 crackers doing oil cracking (fuels too)

Xefyr0:
Screenshots WhitePhant0m shared for posterity:

> "this used to have twice the height"
image

> "that *was* my cracker tower before"
image

> "rest of the base, to the best of my ability"
image

Base was late LuV at the time of peak Cracker/Pyro Oven quantity, using Nitrobenzene for power.

This PR, while well-intentioned, is not going to be a net benefit in my eyes.
To start with, the quantity of Crackers and Pyro Ovens shown in these screenshots does not seem disproportionate to the quantity of machines used in the rest of the base. Furthermore, there are alternative options or planned/implemented changes that will alleviate the concerns that lead to these quantities of machines.

Crackers:
1. LCRs provide the same rate of throughput with LuV power as Crackers would with LuV PCHs, as I noted [here](https://discord.com/channels/1268768059489783838/1523303175749107842/1532585811797741600)
2. Crackers as used in the Transmatrix line is a separate matter - those recipes can and should be considered independently and can be adjusted independently. Crackers performing these recipes were not part of the tower of 64 either.

Pyrolyse ovens:
One of the main bottlenecks of using Nitrobenzene for power is acquiring Benzene in sufficient quantities for powering the base.
1. Wither Gas (#2517 and #2606) would offer a fuel option at this stage of the game that costs significantly less Benzene per unit of energy produced.
2. Planned buffs to turbine rotors in #2674 would offer rotor options with increased fuel efficiency.
3. #2700 would make it easier to make more efficient use of the Benzene reserves of Pyrolyse Oven outputs.

PR #2688, opened 2026-07-19
Change §tTeal to §3Teal in Prismatic Crucible lang entries

DilithiumThoride:
Fixes a server crash when selecting Teal in the Advanced Chroma Detector Hatch.

When changing the Teal color from the standard §3 to the Chroma Colors §t, the MoniLabs Prismatic Crucible code didn't get the memo and so trying to set an Advanced Chroma Detector Hatch to Teal sets the hatch to detect the color `null`, which causes a server crash and also a small (but correctable) bit of savefile corruption.

(The corruption is that if a hatch is saved detecting `null` and is then Loaded, this load will cause all future placed Adv.Chroma Detectors to also be placed detecting `null` not Red. Changing the color of a detector in an invalid state triggers another crash. To fix this, all invalid hatches must be removed and then the server must be restarted.)

### DIFF
--- a/kubejs/assets/monilabs/lang/en_us.json
+++ b/kubejs/assets/monilabs/lang/en_us.json
@@ -99,7 +99,7 @@
   "monilabs.prismatic.color_name.orange": "§6Orange§r",
   "monilabs.prismatic.color_name.pink": "§cPink§r",
   "monilabs.prismatic.color_name.red": "§4Red§r",
-  "monilabs.prismatic.color_name.teal": "§tTeal§r",
+  "monilabs.prismatic.color_name.teal": "§3Teal§r",
   "monilabs.prismatic.color_name.yellow": "§eYellow§r",
   "monilabs.prismatic.current_color": "Current color: %s",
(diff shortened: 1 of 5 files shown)


Xefyr0:
I'm not too experienced in the mods involved in this interaction. (Phoenix's Chromatic Codes & Monilabs)
Requesting review from the SMEs for those mods respectively - is there a way to have our cake and eat it here?

DilithiumThoride:
This may be fixable on the Monilabs side prior to 8.0.0, but will definitely be fixed by 8.0.0 allowing both cake and eating of it. This was my immediate bandaid fix to make it not just automatically crash the game.

(The issues are triggered by an LDLib UI pulling the kjs loc resulting in an inappropriate string comparison with some MoniLabs code that does _not;_ all of which will be inherently fixed when LDLib gets ripped out.)

NegaNote:
It's fixed by [this PR to MoniLabs](https://github.com/Omicron-Industries/MoniLabs/pull/66) already, just needs that to have a cf release.

PR #2696, opened 2026-07-26
nuke configurable maintenance hatch bc it encourages bad habits

NegaNote:


WithersChat (approved):

Xefyr0 (changes requested):
Configurable Maintenance Hatches are mentioned in quest 0523F4AB9A794390.
This needs to be fixed if Configurable Maintenance Hatches are to be removed.

Xefyr0 (approved):

PR #2704, opened 2026-08-01
Xefyr goes on a Chemistry Bender

Xefyr0:
One night I went down like, 5 Wikipedia rabbit holes and made a bunch of chemistry recipes. After a bit of double-checking for balance, they are now in a PR.

These are supposed to improve the pack in several ways, both in realism (Acetic Anhydride fluid->dust, additional recipes for Formic Acid, better Nitrous Acid color) and in gameplay experience. (smelting recipes to reduce ore oxides, new arene conversion recipes, Chloroacetic Mixture moved to Mixer)

I apologize for doing this in such a tangled mess, but the other extreme would be no less than 10 PRs which would not be much of an improvement.


WithersChat:
Requested a jollahs review. Merge if they approve.

PR #2707, opened 2026-08-02
Add lag mitigation for Snad farms in and around lazy loaded chunks

Xefyr0:
When a Snad farm improperly chunkloaded, it can cause a buildup of item entities in block-ticking (a.k.a. "lazy loaded") chunks. Because these sorts of chunks don't tick items, the normal lag mitigation approaches MC has for item entities don't trigger. (Collection of entities into stacks and despawning after 5 mins)
This script was created to prevent item entities that are the product of Snad farms from spawning in these block-ticking chunks to prevent the issue.

This PR was AI-assisted, shared by Mavalle in the #moni-dev discord channel and reviewed/modified by yours truly.


WithersChat:
I approve of this but considering it was AI assisted I think Note should review it too before merging.

NegaNote (approved):
AI-assisted as it may be, the logic is solid.

PR #2709, opened 2026-08-04
Nuke EU P2P from all modes.

AngeliaDusk:
Ever since 0.13, EU P2P has been removed in HM/EM. However, while it is technically in NM, it has an exponential tax, which makes it effectively useless. So, uh...might as well nuke it from existence.
Completely disables EU P2P entirely.

Note for anyone reading the commit names: I accidentally said "Hard Mode" when deleting the Expert Mode config override, just ignore that lmao


Xefyr0 (approved):

kubejs/startup_scripts/_packmode.js
Xefyr0: Missed this line.

Xefyr0 (changes requested):

Xefyr0 (approved):

PR #2711, opened 2026-08-05
Fix GTM fertilizer causing reduced secondary product yield when used in Phytogenic Insolator

### LINKED ISSUE #2710: GTM Fertilizer in Phytogenic Insolator reduces the yield of secondary products
The GTM fertilizer is configured as follows in `kubejs/data/thermal/recipes/machines/insolator`
`{
    "type": "thermal:insolator_catalyst",
    "ingredient": {
        "item": "gtceu:fertilizer"
    },
    "primary_mod": 2.0,
    "secondary_mod":0.5,
    "energy_mod": 0.8,
    "min_chance": 0.5,
    "use_chance": 1.0
}`

secondary_mod is a modifier applied to the yield of a secondary recipe byproduct. since it is configured as zero, recipes that always guarantee a secondary product (such as sapling recipes that produce wood and give back the sapling) will no longer guarantee a secondary byproduct (e.g. with the normal yield of 1.1 saplings per recipe, applying GTM fertilizer drops the yield to 1.1*0.5 = 0.65, which is no longer enough to loop the recipe)

This seems like a very confusing mechanic that is not explained in the quest entry for GTM fertilizer and is not visible in EMI or in the fertilizer item description, and also makes GTM fertilizer unusable for wood automation early game due to net-negative sapling output, which means that the loop will eventually run out of initial saplings and stop.

The most likely solution would be to increase secondary_mod to 1.0, which would result in default yield of secondary product for the recipe, which matches the information shown in EMI.

Archengius:
Fixes #2710.

### DIFF
--- a/kubejs/data/thermal/recipes/machines/insolator/insolator_catalyst_fertilizer.json
+++ b/kubejs/data/thermal/recipes/machines/insolator/insolator_catalyst_fertilizer.json
@@ -4,7 +4,7 @@
         "item": "gtceu:fertilizer"
     },
     "primary_mod": 2.0,
-    "secondary_mod": 1.0,
+    "secondary_mod": 0.95,
     "energy_mod": 0.8,
     "min_chance": 0.5,
     "use_chance": 1.0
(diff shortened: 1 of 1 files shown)


Xefyr0 (approved):

Xefyr0 (changes requested):
Actually, would you change secondary_mod to 0.95 instead?
The point is to reduce secondary outputs, and changing secondary_mod to 0.95 would accomplish that while still being net-positive.

Archengius:
Sure, changed it to 0.95

Xefyr0 (approved):

PR #2713, opened 2026-08-05
update gtmutils to 2.10.1

NegaNote:
updates mechanics for PTERB/WAT:
completely replaces it with two separate multis, the WEB Hub (which keeps the old structure) and the cheaper WEB Receiver

Only the Hub uses coolant if enabled, and coolant has been changed to be multiplied by the log of EU transferred, not a linear relationship


Xefyr0 (approved):
LGTM, but merging this should be postponed as long as possible in the 0.14 dev cycle.
Merging this PR will make it more time-consuming to switch between `main` and `0.14-dev`, since contributors will need to change the version of Monilabs being used on top of switching Git branches.

WithersChat:
Electing to merge this now for ease of developing 0.14, we have reached a point where the benefits outweigh the drawbacks.

PR #2715, opened 2026-08-06
Make PEEK slightly shorter

Xefyr0:
Removes 3 steps from the PEEK line without altering material costs.
Since PEEK is a little bit heavy on the recipe spam, this keeps the core idea of the line while making it less tedious.


WithersChat:
Jollahs in shambles.

(Good change tho.)

PR #2720, opened 2026-08-08
Move to TMRV

Xefyr0:
The same as #2286 but without all of the refactoring, since compat with KJS recipe hiding was implemented in TMRV 0.9.0.

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -626,8 +626,8 @@
       "required": true
     },
     {
-      "fileID": 7391695,
-      "projectID": 238222,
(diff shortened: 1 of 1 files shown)


NegaNote (approved):

NegaNote:
Do we want this to go on the main branch and not 0.14-dev?

Xefyr0:
From what I've seen so far, this doesn't affect much from the players' side of things.
And people shouldn't be mixing different versions on client vs. server anyways.

PR #2725, opened 2026-08-09
Better oreproc guidance

Xefyr0:
One common issue Monifactory players experience is that their ore processing systems are slow or "clog" frequently. Under normal circumstances, this sort of problem is resolved by adding more machines and overclocking those machines to higher voltage tiers. However, these are not normal circumstances.

It turns out that many players are putting every ore through the most intensive ore processing route available: Macerate -> Chem Bath/Ore Washing -> Thermal Centrifuge -> Macerate. Using Bauxite as an example, this would take 140 seconds of machine time total at LV. This quantity absolutely _dwarfs_ the fraction of a second it takes to smelt the raw ore directly in a multi-smelter, and is twice the 71 seconds it takes to perform the Macerate -> Wash -> Forge Hammer -> Centrifuge route suggested in this PR.

When this time-consuming way to process ores is combined with the renewable source of raw ore that Microminers offer, it is no surprise that ore processing is so slow for so many players.

Some have suggested making an "ore processing monolith multiblock" to patch up the consequences of the concerns listed above, but it's hard to agree with a solution that treats the intended behavior of a system (and the lessons it teaches) as a problem to be solved rather than a equilibrium to be maintained.

That's where this PR comes in: By providing alerting players to common pitfalls and informing them which actions to take in response to certain patterns they observe in their ore processing, they will be able to properly engage with and master the system instead of continuing to suffer the punishment for being ignorant of it.


Xefyr0:
That all said, this is no guarantee that this PR will resolve the issue entirely.

Future additions and amendments to the quests in this PR may be necessary, and it may be helpful to increase the duration of most ore-yielding microminer missions as a way to slow down players' ore intake to better match the rates their ore processing systems can handle.

WithersChat:
I mean if it was within the scope of Moni I'd suggest scrapping all oreproc and reworking it from the ground up, but it 100% isn't in scope so this is good enough for 0.14 at least. V8 and its modular multis could allow to make the "oreproc multi" more interesting but that is another discussion.

PR #2728, opened 2026-08-14
move Ender Spores to all modes

PlasmaticVoide:
ender spore recipes in `greenhouse.js` were for all modes so nm had kjs errors
assumed that intended behavior was for spores to be for all modes instead of only hm/em


Xefyr0:
Merging this immediately since it's a straightforward fix for a real issue.

PR #2743, opened 2026-08-26
Microversium-Plated Micro Dimensional Hammer [Tier Three and a Half]

WithersChat:
And its stabilized version, *Stabilized Microversium-Plated Micro Dimensional Hammer [Tier Three and a Half]*.


config/jade/plugins.json
Xefyr0, line 1: Were the changes in this file intentional to include in this PR?

Xefyr0 (approved):
Outside that one point of confusion, LGTM.

Too lazy to boot up an instance with the changes rn but I'm reasonably sure you covered all the bases with this one, partially since it's coming from IM

config/jade/plugins.json
WithersChat: Yes. It's an auto-generated line that happens because pipez exists and so I included it because why the fuck not.

PR #2746, opened 2026-08-28
Makes MAX-Tier Voltage required in several MAX-related recipes

AngeliaDusk:
Makes MAX 2A/4A/16A/64A Dynamo Hatch (WHY WAS THE 4A ONE OpV????????), 4A/16A/64A MAX Energy Hatch, all MAX crafting components, all MAX Laser Hatches, and Causality Exempt Heavy Monic Plating require MAX Energy Hatches.
Well, except for the Fluid Regulator, which was already a MAX recipe.

Intended to make TES more required and less of a random "haha Laser Hatch jumpscare" moment.

Also divides duration for all Laser Hatch recipes by 10 to keep the same times as their UEV and prior siblings.


WithersChat (approved):
Neat. Also, good that 2A MAX energy hatch is still UIV for gating reasons.

NegaNote (approved):

PR #2775, opened 2026-09-16
Slight Naquadah Superfuel and Antimatter Fuel tweaks

AngeliaDusk:
Antimatter Fuel + Neutronium nerfed slightly (1A OpV → 1A UXV) to make the alt recipe more appealing.
Antimatter Fuel + Infinity replaced with Antimatter Fuel + Eltz, and reverting the 16x buff to that recipe (64A MAX back to 4A MAX). Also makes the GTValues more...sane.
Removed Antimatter Fuel + Neutronium from TES outright.
Naquadah Superfuel takes 72L Liquid Eltz now. (Makes it actually uses the fifth fluid slot again too.)

Also, this PR lowers the item input slots in the Naqauadah Refinery from 6 to 3, as slots 4-6 were unused.

**Willing to change anything if needed.**

Note that the Antimatter Collider quest still mentions Infinity. I genuinely cannot be assed to deal with FTB Quests, so someone else can do that.


WithersChat (approved):
This has been requested by me personally in a voice chat.

NegaNote (approved):

PR #2785, opened 2026-09-19
Nuke Redundant Items

crumblbee:
It's always bothered me how there are some useless items in the pack, e.g. ruby rods. This makes things like setting up oreproc unnecessarily annoying, when you check for ruby's recipes. This PR nukes these items.

## Summary of nuked items

- Extended crafting: `redstone_catalyst` and `the_ultimate_component` have no uses.
- GregTech:
    - Many gem plates, e.g. blue topaz plates, have no use.
    - Many gem rods, e.g. amethyst rods, have no use.
    - Activated netherite rods, netherite nuggets, stone rods, stone gears, double graphene plates, have no use.
    - A few dusts, including their small and tiny variants, have no use: Cocoa, netherite, and the five magnetic dusts.
    - A few dusts, including their small and tiny variants, are unobtainable: Perlite, and magnetic terbium dust. 
    - Many of the magnetic components are useless: The ingots (all its products are unmagnetic again!), the nuggets, the screws and the plates.
- ProjectRed's `electrotine_silicon` and `electrotine_silicon_comp` have no apparent use, but can be crafted.

That `gtceu:netherite_dust` has no use strikes me as odd, but I guess that's upstream gregtech. Could add a simple smelting recipe turning it back into netherite.

`itemfilters:strong_nbt` and `itemfilters:weak_nbt` are also uncraftable, but I did not nuke them.

## Method

This was found by dumping all the recipes via the method described in <https://github.com/jackschedel/wikiMC-sources/blob/master/README.md>, and then turning it into a directed graph, with items/fluids serving as nodes, and an edge from A to B existing iff there is a recipe with A as input and B as output. After that, I filtered out macerator recycling recipes, arc furnace recycling recipes, and collapsed the strongly connected components of the graph. The program then printed all the leaves and roots of the graph, i.e. items that are unobtainable, or items that have no use.

This list then had to be filtered harshly by hand, e.g. filtering out building blocks (which obviously have use), and compared against the actual ingame data. This took a lot of time.

I'm only mostly sure that the nukelist is the appropriate place for all these.

## Things I decided against nuking

I did not nuke all the different stone ores, though they do clutter up EMI a lot, especially when you look for "what smelts into copper ingots?". I'm guessing this would have to go on the client-side EMI nuke script anyway?

I didn't nuke anything about nuclearcraft, because I don't understand the mod well and it's going to be removed soon anyway.

Unrelated to the PR, but GT fluid cells are a noob trap and could stand to be nuked.


crumblbee:
Closing in favour of [this EMI resourcepack](https://github.com/crumblbee/gregtech-cleaner-emi), which already does all this.
If any parts of the resourcepack should be integrated into moni itself, e.g. making acronyms like "EBF" searchable in EMI, let me know and I can see if I can submit a PR along those lines.

Cosmoros-git:
Here to comment on that resource pack. The issue with it is that if you use it early in the game, it breaks EMI crafting. It's nice that the recipes are hidden, but it's only a midgame at least resouce pack.

PR #2787, opened 2026-09-20
Add interactive install.sh dedicated server installer

imdonix:
Adds a single interactive `install.sh` script that automates the manual
dedicated-server setup from the README, plus documents it in a new
"Automatic setup" section.


WithersChat:
This isn't really useful or worth the extra maintenance effort of keeping it working across updates that might break it.

(Also, I do not know whether or not this is AI assisted, but for future reference any AI assistance needs to be disclosed. If this isn't AI assisted please ignore that last part.)

imdonix:
Fair enough! I feel like the manual setup is a bit clunky for beginners when trying to quickly spin up a server, but I get that maintaining it takes effort so I see where you're coming from.

(FYI, I just adapted this script from another modpack template for Moni, so idk its AI or not)

